Issue #172, opened 2019-05-20
Is a DSVM required?


ctstone:
Consider documenting the steps to run the notebooks on a vanilla Ubuntu VM

ctstone:
@jiata should we close this issue? I see a PR was linked and merged...

Issue #171, opened 2019-05-20
Provide az cli command to create ideal VM


ctstone:
Recommend providing an `az` CLI snippet to create the ideal VM with minimum SKU and desired image name (`az vm create ... `)

PR #222, opened 2019-06-05
[#171 & #172] dsvm tool

jiata:
#171 & #172

When we make the repo public, we wont need auth to clone the git repo, so we can add functionality that will install the latest bits on master to the machine.


PatrickBue:
@jiata Fyi, as you know, there will be a /contrib/ folder where un-tested possible non-maintained code will live. Once the folder is there, could move your tool to e.g.  /contrib/vm_creation_script/

PR #292, opened 2019-08-14
training intro notebook for detection

jiata:
### Description
training notebook + utils for it.
ground work for webcam notebook to build off from

### Checklist:

- [X] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.


jiata:
currently merging into stagingOD... ideally should merge into `staging` but might have more conflicts if i do so.

environment.yml
```diff
@@ -18,7 +18,8 @@ channels:
 dependencies:
 - python==3.6.8
 - pytorch==1.0.0
-- torchvision  
+- torchvision>=0.3.0
+- pycocotools>=2.0
```
PatrickBue: Not sure if pycocotools exists for windows. If not, I used this here before (but not sure this is the best way):
pip install "git+https://github.com/philferriere/cocoapi.git#egg=pycocotools&subdirectory=PythonAPI"

utils_cv/detection/data.py
```diff
@@ -0,0 +1,119 @@
...
+        "sheep",
+        "cow",
+        "elephant",
+        "bear",
+        "zebra",
+        "giraffe",
+        "N/A",
+        "backpack",
+        "umbrella",
+        "N/A",
```
PatrickBue: Are the N/A's expected?

utils_cv/detection/data.py
```diff
@@ -0,0 +1,119 @@
...
+    # for now hardcoding base url into Urls class
+    base = (
+        "https://cvbp.blob.core.windows.net/public/datasets/object_detection/"
+    )
+
+    # traditional datasets
+    fridge_objects_path = urljoin(base, "odFridgeObjects.zip")
+
+
+def get_coco_labels() -> List[str]:
```
PatrickBue: Maybe just call coco_labels(), similar to the imagenet_labels() function for classification

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+import torch
+from torch.utils.data import Dataset, Subset
+import xml.etree.ElementTree as ET
+from PIL import Image
+
+from .plot import display_bounding_boxes, plot_grid
+from .model import get_transform
+
+
+class DetectionDataset(object):
```
PatrickBue: - If you copied (and slightly edited) this code from somewhere, can you provide a link. 
- We should brainstorm how this code here (copied and edited) can be best maintained. E.g. should we change the code to look nicer? Should we remove functions which we don't use, should we combine/refactor some of the functions (maybe not since that'll make it harder for us to copy newer updates of the code from the source in the future), etc.
- [minor and unsure] Should we add this functionality to data.py rather than having two .py files? classification does not have a dataset.py, maybe by putting everything into data.py we are more aligned?

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Raises:
+            Exception is idx is not None and rand is set to True
+
+        Returns:
+            A tuple of boxes, categories, image path
+        """
+        if (idx is not None) and (rand is True):
+            raise Exception("idx cannot be set if rand is set to True.")
+        if idx is None and rand is False:
```
PatrickBue: Probably better to throw an error here that rand needs to be explicitly set to true if idx is not specified.

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+        self.annotation_folder = annotation_folder
+
+        self.ims = list(sorted(os.listdir(self.root / self.image_folder)))
+        self.annotations = list(
+            sorted(os.listdir(self.root / self.annotation_folder))
+        )
+        self.categories = self._get_categories()
+
+    def _get_categories(self) -> List[str]:
+        """ Parses all Pascal VOC formatted annoatation files to extract all
```
PatrickBue: [minor] typo: annoatation

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+            rand = True
+        if rand:
+            idx = randrange(len(self.ims))
+        boxes, labels, im_path = self._get_im_data(idx)
+        return (boxes, [self.categories[label] for label in labels], im_path)
+
+    def split_train_test(
+        self, train_ratio: float = 0.8
+    ) -> Tuple[Dataset, Dataset]:
+        """ Split theis dataset into a training and testing set
```
PatrickBue: [minor] typo: theis

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Args:
+            train_ratio: the amount of images to use for training (the rest
```
PatrickBue: [minor] not amount but ratio

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+            will be used for testing.
+
+        Return
+            A training and testing dataset in that order
+        """
+        test_num = math.floor(len(self) * (1 - train_ratio))
+        indices = torch.randperm(len(self)).tolist()
+        self.set_transform(get_transform(train=True))
```
PatrickBue: Don't fully understand this line, e.g. does the second call to self.set_transform() over-write the first call and hence the first call has no real effect? Probably not, but please make sure this self.set_transform() has the desired effect.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+from torchvision.models.detection.faster_rcnn import FastRCNNPredictor
+from torch.utils.data import Dataset, DataLoader
+import matplotlib.pyplot as plt
+
+from .references.transforms import RandomHorizontalFlip, Compose, ToTensor
+from .references.engine import train_one_epoch, evaluate
+from .references.utils import collate_fn
+from .references.coco_eval import CocoEvaluator
+from .plot import BoundingBoxParams, plot_boxes, plot_grid
+
```
PatrickBue: Same comment as before: if you copied (and edited) this code from somewhere, can you please provide a link.

utils_cv/detection/plot.py
```diff
@@ -0,0 +1,279 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
```
PatrickBue: Same comment as before: if you copied (and edited) this code from somewhere, can you please provide a link.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+from typing import List, Tuple, Union
+from pathlib import Path
+from functools import partial
+
+from PIL import Image
+import cv2
```
PatrickBue: Any chance we can avoid using cv2? I have seen problems before getting opencv running on a webservice. Also, opencv's imread read in images in BGR order, all other packages in RGB, this is confusing and will almost definitely lead to silent bugs.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+            ax: axes to draw on
+
+        Returns nothing but plots graph.
+        """
+        im_path = (
+            self.dataset.root
+            / self.dataset.image_folder
+            / self.dataset.ims[idx]
+        )
+        im = cv2.imread(str(im_path))
```
PatrickBue: Looks like we can get rid of cv2 by loading using PIL here instead? Out of interest, the ax.imshow() does that show the image in flipped RGB order?

utils_cv/detection/plot.py
```diff
@@ -0,0 +1,279 @@
...
+Helper module for visualizations
+"""
+from typing import List, Union, Tuple, Callable, Any
+from pathlib import Path
+
+import cv2
```
PatrickBue: Let's try to get rid of the cv2 here as well

utils_cv/detection/plot.py
```diff
@@ -0,0 +1,279 @@
...
+        im: The image to plot boxes on
+        boxes: the boxes to plot - a list of [xmin, ymin, xmax, ymax] boxes
+        categories: list of catogires to label the boxes with
+        bbox_params: the parameter of the bounding boxes
+
+    Returns:
+        The same image with boxes and labels plotted on it
+    """
+    # Convert to RGB
+    im = cv2.cvtColor(im, cv2.COLOR_BGR2RGB)
```
PatrickBue: I know this isn't your code, but just to highlight my comment earlier regarding opencv and the silent RGB vs BGR errors: this is ugly. Whoever wrote this assumed that the input image was a numpy array in BGR order, and not in RGB order.

utils_cv/detection/plot.py
```diff
@@ -0,0 +1,279 @@
...
+            rect_th,
+            rect_color,
+            text_size,
+            text_th,
+            text_color,
+        )
+
+
+def plot_boxes(
+    im: JpegImageFile,
```
PatrickBue: Never seen this before "JpegImageFile". Maybe this is the right thing to do, maybe it's over-kill: im is just a numpy array albeit of (w x h x c) with c = {1,3}

PatrickBue:
@jiata This is a great start, and looks very good already!!!

utils_cv/detection/dataset.py
jiata: i did wrote this class, didn't copy it. It is loosely based on this: https://pytorch.org/tutorials/intermediate/torchvision_tutorial.html#writing-a-custom-dataset-for-pennfudan (which i will add as reference in a comment)

utils_cv/detection/model.py
jiata: i wrote this class

utils_cv/detection/plot.py
jiata: wrote these funcs

utils_cv/detection/plot.py
jiata: yeah, its a cv2 thing, if we switch to pillow we just use the np array then.

utils_cv/detection/dataset.py
jiata: also, all the functions in this class are used

utils_cv/detection/dataset.py
jiata: will refactor this, as you point out, this is a bit sloppy.

utils_cv/detection/model.py
jiata: yeah, i can update to use pillow, forgot that cv2 is a pain sometimes.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+    qualified_labels = []
+    qualified_boxes = []
+    for label, box, score in zip(pred_labels, pred_boxes, pred_scores):
+        if score > threshold:
+            qualified_labels.append(label)
+            qualified_boxes.append(box)
+    return qualified_labels, qualified_boxes
+
+
+def get_transform(train: bool) -> List[object]:
```
jiata: this is a snippet from the torchvision tutorial: https://pytorch.org/tutorials/intermediate/torchvision_tutorial.html#writing-a-custom-dataset-for-pennfudan

environment.yml
jiata: https://github.com/cocodataset/cocoapi/issues/169#issuecomment-462528628
it might be the best way

utils_cv/detection/data.py
jiata: not sure.. but i copied from the official pytorch docs: https://github.com/pytorch/vision/blob/master/docs/source/models.rst#object-detection-instance-segmentation-and-person-keypoint-detection

utils_cv/detection/dataset.py
jiata: WRT to moving this functionality to data.py -- this is where i had initially written it in, but it got too big and unwieldy, which was why i moved it to a separate file.

PatrickBue: Thanks for the info!

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+            image_folder: the name of the image folder
+            annotation_folder: the name of the annotation folder
+        """
+
+        self.root = Path(root)
+        self.transforms = transforms
+        self.image_folder = image_folder
+        self.annotation_folder = annotation_folder
+
+        self.ims = list(sorted(os.listdir(self.root / self.image_folder)))
```
PatrickBue: Maybe rename to im_paths to make more clear what ims is

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+        self.annotations = list(
+            sorted(os.listdir(self.root / self.annotation_folder))
+        )
+        self.categories = self._get_categories()
+
+    def _get_categories(self) -> List[str]:
+        """ Parses all Pascal VOC formatted annoatation files to extract all
+        possible categories. """
+        categories = ["__background__"]
+        for annotation_path in self.annotations:
```
PatrickBue: I personally like the syntax: for NAME in NAMES which I believe is also often used in the other code. Ie. here could be: for anno_path in anno_paths

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+            labels.append(self.categories.index(category.text))
+
+        # get image path from annotation
+        annotation_dir = os.path.dirname(annotation_path)
+        im_path = root.find("path").text
+        im_path = os.path.realpath(os.path.join(annotation_dir, im_path))
+
+        return (boxes, labels, im_path)
+
+    def _get_im_data(self, idx) -> Tuple[List[List[int]], List[int], str]:
```
PatrickBue: Since this function does nearly the same as "self._get_annotations", but takes an index rather than a path as input, should it maybe be called similar. E.g. get_anno_path() and get_anno_idx()

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+        """ Show batch of images.
+
+        Args:
+            rows: the number of rows images to display
+
+        Returns None but displays a grid of annotated images.
+        """
+        plot_grid(display_bounding_boxes, self.get_image_features, rows=rows)
+
+    def _get_annotations(
```
PatrickBue: This function returns multiple rectangles, and for each the coordinates (boxes), labels, and path. I found it very helpful in the past to have a class called BBox or something (but not rectangle, because that means something else in OpenCV). That way one can just cal box.score, box.label, box.rectangle(), box.upperLeftCorner(), box.width(), etc. Make it much easier to work with and also leads to much nicer code. See here for an example implementation:
https://github.com/Azure/ObjectDetectionUsingCntk/blob/master/helpers.py#L943

We would use this bbox object not just for ground truth annotation, but also when running inference the DNN would return a list of bboxes. That helped me a lot in the past, e.g. I was always confused which are the upper-left corner coordinates, which are the lower-right corner coordinates in just the "boxes" variable.

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+            tree = ET.parse(annotation_path)
+            root = tree.getroot()
+            objs = root.findall("object")
+            for obj in objs:
+                category = obj[0]
+                assert category.tag == "name"
+                categories.append(category.text)
+        return list(set(categories))
+
+    def get_image_features(
```
PatrickBue: "features" is unclear and a bit misleading. Maybe name get_annos() or get_anno_bboxes()   (see comment below what I mean with the BBox class)

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        test_num = math.floor(len(self) * (1 - train_ratio))
+        indices = torch.randperm(len(self)).tolist()
+        self.set_transform(get_transform(train=True))
+        train = Subset(self, indices[:-test_num])
+        self.set_transform(get_transform(train=False))
+        test = Subset(self, indices[-test_num:])
+        return train, test
+
+    def set_transform(self, transforms: List[object]) -> None:
```
PatrickBue: Do we need this function? If it is such that the naming is similar to fastai then I understand. It's just that transforms is public and could be set directly, this function isn't doing anything on top.

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        Returns
+            (boxes, labels, im_path)
+        """
+        annotation_path = (
+            self.root / self.annotation_folder / str(self.annotations[idx])
+        )
+        return self._get_annotations(annotation_path)
+
+    def __getitem__(self, idx):
```
PatrickBue: I don't fully understand how your split into train_test function works. But I guess this would only iterate over the training or test elements (if the objects representing the splits are being used)?

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+        Returns:
+            A tuple of boxes, categories, image path
+        """
+        if (idx is not None) and (rand is True):
+            raise Exception("idx cannot be set if rand is set to True.")
+        if idx is None and rand is False:
+            rand = True
+        if rand:
+            idx = randrange(len(self.ims))
+        boxes, labels, im_path = self._get_im_data(idx)
```
PatrickBue: Is label here an integer, and category the actual class name? Is that always the case? Nice point for a BBox class, there we could have a "class_name" and "class_id" fields

utils_cv/detection/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+        train = Subset(self, indices[:-test_num])
+        self.set_transform(get_transform(train=False))
+        test = Subset(self, indices[-test_num:])
+        return train, test
+
+    def set_transform(self, transforms: List[object]) -> None:
+        """ Apply transformations. """
+        self.transforms = transforms
+
+    def show_batch(self, rows: int = 1) -> None:
```
PatrickBue: Does this show a batch or simply shows some images (probably starting from first image in list)

utils_cv/detection/plot.py
```diff
@@ -0,0 +1,279 @@
...
+                cv2.FONT_HERSHEY_SIMPLEX,
+                bbox_params.text_size,
+                color=bbox_params.text_color,
+                thickness=bbox_params.text_th,
+            )
+
+    return im
+
+
+def display_bounding_boxes(
```
PatrickBue: Nice example where we'd just have to pass in a list of BBox objects, and each BBox contains all the necessary information

utils_cv/detection/plot.py
```diff
@@ -0,0 +1,279 @@
...
+def _setup_pr_axes(ax: plt.axes, title: str) -> plt.axes:
+    """ Setup the plot settings for plotting PR curves. """
+    ax.set_xlabel("recall", fontsize=12)
+    ax.set_ylabel("precision", fontsize=12)
+    ax.set_xlim(0, 1.0)
+    ax.set_ylim(0, 1.01)
+    ax.set_title(title, fontsize=14)
+    ax.grid(True)
+    return ax
+
```
PatrickBue: Could add logical separator here that everything below is for Precision-Recall curve plotting. E.g.

# ===== Precision - Recall curve =====

utils_cv/detection/plot.py
```diff
@@ -0,0 +1,279 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
PatrickBue: General though about plotting, and don't know the answer: maybe there is a common library we could use for plotting, etc, rather than having to implement it ourselves. Something to keep in mind going forward..

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+import cv2
+import torch
+import torch.nn as nn
+from torchvision import transforms
+from torchvision.models.detection import fasterrcnn_resnet50_fpn
+from torchvision.models.detection.faster_rcnn import FastRCNNPredictor
+from torch.utils.data import Dataset, DataLoader
+import matplotlib.pyplot as plt
+
+from .references.transforms import RandomHorizontalFlip, Compose, ToTensor
```
PatrickBue: Not sure if "references" is the best name. Bet let's keep for now, can always change later.

utils_cv/detection/plot.py
```diff
@@ -0,0 +1,279 @@
...
+
+import cv2
+import numpy as np
+import matplotlib.pyplot as plt
+from PIL.JpegImagePlugin import JpegImageFile
+
+from .references.coco_eval import CocoEvaluator
+
+
+class BoundingBoxParams:
```
PatrickBue: Name does not make clear that these params have to do with plotting. Maybe use PlotSettings?

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+from .references.engine import train_one_epoch, evaluate
+from .references.utils import collate_fn
+from .references.coco_eval import CocoEvaluator
+from .plot import BoundingBoxParams, plot_boxes, plot_grid
+
+
+def get_bounding_boxes(
```
PatrickBue: It's unclear from the fct name what the fct does

utils_cv/detection/model.py
PatrickBue: Maybe this function should be in dataset? It's only used there atm

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+        lr: float,
+        train_ratio: float = 0.8,
+        batch_size: int = 2,
+        model: nn.Module = None,
+        momentum: float = 0.9,
+        weight_decay: float = 0.0005,
+    ):
+        """ Initialize leaner object. """
+        self.device = (
+            torch.device("cuda")
```
PatrickBue: Could make a helper function out of this called e.g. "def torch_device()" if this phrase is used more often.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+            if torch.cuda.is_available()
+            else torch.device("cpu")
+        )
+        self.model = model
+        self.dataset = dataset
```
PatrickBue: As we already discussed, would be nicer if this takes an already split into train and test dataset as input.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+        ax.set_ylabel("loss")
+        ax.plot(self.losses)
+
+    def evaluate(self, dl: DataLoader = None) -> CocoEvaluator:
+        """ eval code on validation/test set. """
+        if dl is None:
+            dl = self.test_dl
+        return evaluate(self.model, dl, device=self.device)
+
+    def get_model(self) -> nn.Module:
```
PatrickBue: Do we need this? Can just write self.model instead of self.get_model()

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+        """ returns the model object. """
+        return self.model
+
+    def inference(self, im_path: Union[str, Path]):
```
PatrickBue: Could rename this to predict() to be aligned with fastai. Then the input could be "im_or_path" with a type check if it is a path, then read image, otherwise im_or_path is already the image.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+        """ Performs inferencing on an image path. """
+        im = Image.open(im_path)
+        transform = transforms.Compose([transforms.ToTensor()])
+        im = transform(im).cuda()
+        model = self.get_model().eval()  # eval mode
+        with torch.no_grad():
+            pred = model([im])
+        return pred
```
PatrickBue: Could first create a list of BBox objects, and then return that. That's so that we only ever return/input BBox objects and avoid movind around coordinates, labels, etc.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,283 @@
...
+        )
+        boxes, categories, im_path = self.dataset.get_image_features(idx)
+        im = plot_boxes(im, boxes, categories, bbox_params=ground_truth_params)
+
+        # show image
+        ax.set_xticks([])
+        ax.set_yticks([])
+        ax.imshow(im)
+
+    def show_preds(self, ds: Dataset = None, rows: int = 1) -> None:
```
PatrickBue: [minor] Maybe this is the right thing to do, having eg a helper function called plot_grid. I might myself though repeatedly thinking that this looks too complicated and is not easy to follow. I also somehow don't like the fact that the plotting function executes the model. Seems somehow that this function is doing too much, but could be wrong.

PatrickBue:
@jiata Hi JS! As you can see I left a few more comments. Again, this is an amazing starting point, and after some fine-tuning this is ready to go. My main feedback point is to create a BBox class like the one in the link below to store all bounding box specific information such as co-ordinates, ground truth label, potentially also predicted label and predicted score. That way I found in the past that I was able to significantly clean up the code and maybe more importantly avoid bugs (e.g. having to remember in the co-ordinates which one is which, or if they are relative or absolute etc)

https://github.com/Azure/ObjectDetectionUsingCntk/blob/master/helpers.py#L943

environment.yml
simonzhaoms: @jiata I got incompatible packages:
```console
$ conda env create -f environment.yml -n cvbp
Collecting package metadata (repodata.json): done
Solving environment: | 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed                                                                                                                                                                                                             
                                                                                                                                                                                                                   
UnsatisfiableError: The following specifications were found to be incompatible with each other:                                                                                                                    



Package pytorch conflicts for:
pytorch==1.0.0
fastai==1.0.48 -> pytorch[version='>=1.0.0']
torchvision[version='>=0.3.0'] -> pytorch[version='1.2.0|1.2.0+cu92|>=1.1.0']
fastai==1.0.48 -> torchvision -> pytorch[version='1.2.0|1.2.0+cu92|>=0.3|>=0.4|>=1.1.0']
```

utils_cv/detection/dataset.py
jiata: yes

utils_cv/detection/dataset.py
jiata: yes, i've been using label to indicate int and category to indicate classname, bbox class wuodl help for sure.

utils_cv/detection/dataset.py
jiata: actually i just show a random set of images. i use the dataset object (not the dataloader), so no mini_batch here. it might be nicer to use the dataloader and show mini_batch. I think that's what fastai does... but not really sure its that important, the point is just to show some images to make sure the user can see that the annotations are correct. maybe i'll rename to show_ims for now.

utils_cv/detection/model.py
jiata: yeah, i'll make it so that the plotting function doesn't execute the function. but i find the `plot_grid` util to be nice because multiple functions use it and we can reduce the repeated code.

utils_cv/detection/dataset.py
PatrickBue: Make sense. Adding a show_batch() is something we could add in the future. It's nice to see/check that the selected augmentations do the right thing (e.g. they look reasonable, and also rectangles are being flipped if the image is flipped, etc.)

utils_cv/detection/dataset.py
PatrickBue: Vanja pointed out to me sometime ago that we use label/class inconsistenly in classification. Her point is that "classes" should refer to the set of distinct concepts ie classes = set("dog", "cat", "person") while label is an instation, e.g. for a given image where 5 objects were found the returned labels could be labels=["dog", "dog", "cat", "person", "cat"].
So given that, should we use e.g. "labels" if its the names (which is the more common case I am guess), and "label_ids" if its their index?

utils_cv/detection/helper.py
```diff
@@ -0,0 +1,128 @@
...
+            return False
+        if (
+            min(self.rect()) < -self.MAX_VALID_DIM
+            or max(self.rect()) > self.MAX_VALID_DIM
+        ):
+            return False
+        return True
+
+
+class Annotation:
```
PatrickBue: This seems like a good case where could derive from BBox and call AnnotationBBox, and then also have a PredictedBBox.

utils_cv/detection/helper.py
```diff
@@ -0,0 +1,128 @@
...
+# Licensed under the MIT License.
+
+from typing import List, Union
+
+
+class Bbox:
+    """ Util to represent bounding boxes
+
+    Source:
+    https://github.com/Azure/ObjectDetectionUsingCntk/blob/master/helpers.py
```
PatrickBue: Feel free to edit/remove content etc as you see fit.

utils_cv/detection/helper.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
PatrickBue: The name "Helper.py" seems a bit too generic. Maybe bbox.py instead for everything bbox related?

utils_cv/detection/helper.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class Annotation:
+    """ Contains a Bbox. """
+
+    def __init__(
+        self, bbox: Bbox, category_idx: int, category_name: str = None
+    ):
+        self.bbox = bbox
+        self.category_idx = category_idx
```
PatrickBue: To be consistent with "classification", should be call it "label" instead of category?

utils_cv/detection/dataset.py
```diff
@@ -47,21 +68,21 @@ def __init__(
...
-            image_folder: the name of the image folder
+            im_folder: the name of the image folder
             annotation_folder: the name of the annotation folder
         """
 
         self.root = Path(root)
         self.transforms = transforms
-        self.image_folder = image_folder
-        self.annotation_folder = annotation_folder
+        self.im_folder = im_folder
```
PatrickBue: Do we use "folder" or "dir" for classification? I found one example for "dir" for now, but not sure that applies generally to our utils. Ideally this here should be following the example what is already in utils_cv.

utils_cv/detection/model.py
```diff
@@ -100,18 +86,15 @@ def __init__(
...
-        )
+        self.device = torch_device()
         self.model = model
         self.dataset = dataset
         self.lr = lr
         self.momentum = momentum
         self.weight_decay = weight_decay
 
         # get training dataloaders and datasets
+        self.dataset = dataset
```
PatrickBue: I believe the line self.dataset=dataset is a duplicate. Also, should we remove setting train_ds, test_ds and train_dl since we can refer to them using the dataset object. That might also prevent confusion going forwards if somebody updates self.dataset but forgets to also update self.train_ds

PatrickBue:
Thanks for making the changes, def looking better and better! I had a quick look at the newest updates and left some comments. Let me know once you're ready and happy to do a more in-depth scan.

utils_cv/detection/helper.py
jiata: yep, made some updates, added a few functions, hesitant to delete the un-used ones at this moment thought because they look useful. only a few anyway.

utils_cv/detection/model.py
```diff
@@ -167,109 +185,112 @@ def fit(self, epochs: int, print_freq: int = 10) -> None:
...
 
-        plot_grid(
-            self.show_detection_vs_ground_truth,
-            partial(next, self.ds_iterator),
-            rows=rows,
-        )
+                im_idx = int(info["image_id"].numpy())
+                im_path = dl.dataset.dataset.get_path_from_idx(im_idx)
+
+                anno_bboxes = get_bounding_boxes(
```
PatrickBue: Technically this is not an annotation bbox, but a prediction bbox. Maybe just rename to bboxes since it's clear in this context?

utils_cv/detection/model.py
```diff
@@ -167,109 +185,112 @@ def fit(self, epochs: int, print_freq: int = 10) -> None:
...
-        pred_labels, pred_boxes = get_bounding_boxes(pred)
-        pred_params = PlotSettings(rect_color=(255, 0, 0), text_size=0)
-        im = plot_boxes(
-            im,
-            pred_boxes,
-            [self.dataset.categories[l] for l in pred_labels],
-            plot_settings=pred_params,
-        )
+        pred_generator = self.predict_batch(dl, threshold=threshold)
+        detections = [pred for preds in pred_generator for pred in preds]
```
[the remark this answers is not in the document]
PatrickBue: Maybe rename "detections" to "bboxes" since that is more clear. Or a different name as long as its consistent in the code (e.g. predictions vs. detections: think I have seen both used)

utils_cv/detection/plot.py
```diff
@@ -37,30 +39,32 @@ def __init__(
 
 
 def plot_boxes(
-    im: JpegImageFile,
-    boxes: List[List[int]],
-    categories: List[str],
+    im: PIL.Image.Image,
+    anno_bboxes: List[AnnotationBbox],
```
PatrickBue: If this function also works for detected bounding boxes then I would rename "anno_bboxes" maybe into simply "bboxes". Same comment also applies to other places in this file.

utils_cv/detection/plot.py
```diff
@@ -37,30 +39,32 @@ def __init__(
...
+    im_title: str = None,
```
PatrickBue: [minor] could just rename to "title" rather than "imtitle"

utils_cv/detection/plot.py
```diff
@@ -116,6 +131,98 @@ def display_bounding_boxes(
...
+                plot_func(arguments, ax)
+            except Exception:
+                plot_func(*arguments, ax)
+
+    plt.subplots_adjust(top=0.8, bottom=0.2, hspace=0.1, wspace=0.2)
+
+
+def plot_detection_vs_ground_truth(
+    im_path: str,
+    detection_annos: List[AnnotationBbox],
```
PatrickBue: Not sure those two variable names are consistent with your other code: "detection_annos" and "ground_truth_annos". Maybe rename to annno_bboxes and pred_bboxes but not sure right now this is how to call them.

utils_cv/detection/plot.py
```diff
@@ -116,6 +131,98 @@ def display_bounding_boxes(
...
+    # show image
+    ax.set_xticks([])
+    ax.set_yticks([])
+    ax.imshow(im)
+
+
+# ===== Precision - Recall curve =====
+
+
 def _get_cmap(n, name="hsv"):
```
PatrickBue: At the moment this function is only used once, hence not sure the function is needed.

utils_cv/detection/plot.py
jiata: mm i see.. i think i've been thinking of the word 'annotation' a bit differently; like a prediction can produce a set of annotations. But i think you're right about this being confusing... the word "annotation" should strictly mean labelled data, in which case we'll need another word to represented predicted data. I'll go through the code base and make this update. Sorry for this confusion.

utils_cv/detection/model.py
```diff
@@ -167,109 +185,112 @@ def fit(self, epochs: int, print_freq: int = 10) -> None:
...
+        return detections
 
-        # plot ground truth boxes
-        ground_truth_params = PlotSettings(rect_color=(0, 255, 0), text_size=0)
-        boxes, categories, im_path = self.dataset.get_image_features(idx)
-        im = plot_boxes(
-            im, boxes, categories, plot_settings=ground_truth_params
-        )
+    def predict_batch(
```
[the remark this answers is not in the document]
jiata: Actually i think i disagree here. If for example, a user created a DataLoader that has thousands of images, they may not want to deal with all the images at once. And so a prediction generator may be more useful (potentially also from a memory perspective).

utils_cv/detection/bbox.py
```diff
@@ -0,0 +1,139 @@
...
+        self.label_idx = label_idx
+        self.label_name = label_name or str(label_idx)
+
+    @classmethod
+    def from_array(
+        cls, arr: List[int], label_idx: int, label_name: str = None
+    ) -> "AnnotationBbox":
+        """ Create a Bbox object from an array [left, top, right, bottom] """
+        bbox = super().from_array(arr)
+        bbox.__class__ = AnnotationBbox
```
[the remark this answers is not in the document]
jiata: When i call `bbox = super().from_array(arr)`, the bbox object is type `Bbox`. After calling `bbox.__class__ = AnnotationBbox`, we cast the bbox object to an AnnotationBbox, so the type will be `AnnotationBbox`. So, that line is needed.

I'm not totally sure what you mean by your second question: _Also, what about a type(regular_bbox) does that return the type BBox or do we need to set bbox.class = Bbox in the BBox constructor?_

utils_cv/detection/plot.py
jiata: If it makes sense, i'll implement it so that DetectionBbox inherits from AnnotationBbox which inherits from Bbox. 

I think the more logically intuitive implementation is that both DetectionBbox and AnnotationBbox inherit from Bbox, but practically, the DetectionBbox shoudl contain everything that the AnnotationBbox contains. I'll also make Bbox a 'private' class, so to speak, prefixing it's name with and underscore: `_Bbox`. This is to indicate that only the DetectionBbox or AnnotationBbox be used.

utils_cv/detection/model.py
```diff
@@ -0,0 +1,251 @@
...
+        model = self.model.eval()
+
+        for i, batch in enumerate(dl):
+            ims, infos = batch
+            ims = [im.cuda() for im in ims]
+            with torch.no_grad():
+                preds = model(list(ims))
+
+            bbox_batch = []
+            for pred, info in zip(preds, infos):
```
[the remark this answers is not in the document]
jiata: so predict_dl returns predictions independent of batch size... Some thoughts on why we should keep predict_dl and predict_batch in a comment below

utils_cv/detection/model.py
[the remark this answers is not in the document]
jiata: do you have a suggestion for the long syntax? afaik it is the shortest/cleanest way to do a nested list comprehension.

utils_cv/detection/bbox.py
```diff
@@ -139,4 +144,46 @@ def __repr__(self):
...
+
+    @classmethod
+    def from_array(
+        cls, arr: List[int], score: float, **kwargs
+    ) -> "DetectionBbox":
+        """ Create a Bbox object from an array [left, top, right, bottom]
+        This funciton must take in a score.
+        """
+        bbox = super().from_array(arr, **kwargs)
+        bbox.__class__ = DetectionBbox
```
PatrickBue: Should this line "bbox.__class__ = DetectionBbox" be in the constructor? Same comment applies also to the AnnotationBbox code

utils_cv/detection/model.py
```diff
@@ -217,21 +198,22 @@ def evaluate(self, dl: DataLoader = None) -> CocoEvaluator:
     def predict(
         self,
         im_or_path: Union[np.ndarray, Union[str, Path]],
-        threshold: int = 0.6,
-    ) -> List[AnnotationBbox]:
+        threshold: Optional[int] = 0.6,
```
PatrickBue: [minor] Please change here to 0.5 as well as predict_dl

utils_cv/detection/model.py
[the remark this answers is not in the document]
PatrickBue: Sounds good!

utils_cv/detection/plot.py
```diff
@@ -15,7 +15,7 @@
 import matplotlib.pyplot as plt
 
 from .references.coco_eval import CocoEvaluator
-from .bbox import AnnotationBbox
+from .bbox import _Bbox, AnnotationBbox, DetectionBbox
```
PatrickBue: [minor] Should Bbox be private or not? Ie. since it is used here in this fct, do we think the user will also need it?

utils_cv/detection/model.py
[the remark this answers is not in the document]
PatrickBue: I am likely wrong: would something like this here also work? 
     detections =list(pred_generator)

PR #293, opened 2019-08-14
Test updates to match Patricks PR 278

bethz:
### Description
Added support for marker Patrick added: Linux-gpu

### Related Issues
#278 

### Checklist:

### DIFF
--- a/.ci/AML-nightly-gpu.yml
+++ b/.ci/AML-nightly-gpu.yml
@@ -22,7 +22,7 @@ variables:
 
 
 schedules:
-- cron: "56 22 * * *"
+- cron: "30 22 * * *"
   displayName: Daily run
   branches:
     include:
@@ -48,7 +48,7 @@ jobs:
   - name  : 'junitxml'
     value : 'reports/test-smoke.xml'
   - name  : 'testmarkers'
-    value : '"not notebooks and not spark and not gpu and not azuremlnotebooks"'
+    value : '"not notebooks and not spark and linuxgpu and not azuremlnotebooks"'
   - name  : 'expname'
     value : 'nightly-smoke-test-linux-gpu'
 
--- a/.ci/AML-nightly-gpu.yml
+++ b/.ci/AML-nightly-gpu.yml
@@ -1,25 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 #
-variables:
-- group: AzureMLConfig
-- name  : 'test'
-  value : '.ci/run_pytest.py'
-- name  : 'maxnodes'
-  value : 4
-- name  : 'clustername'
-  value : 'cv-nightly-gpu'
-- name  : 'vmsize'
-  value : 'STANDARD_NC6'
-- name  : 'dockerproc'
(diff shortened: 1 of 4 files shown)


.ci/AML-nightly-gpu.yml
```diff
@@ -37,20 +18,147 @@ pr:
...
 jobs:
-- job: Smoke
-  displayName : 'Smoke: no CV tests yet'
+- job: SmokeCPU
+  displayName : 'SmokeCPU'
   pool:
     vmImage: 'ubuntu-16.04'
 
   variables:
+  - group: AzureMLConfig
```
jiata: can these be parameterized with yaml templating or something? there is a lot of duplicate code with the variables. It would be a bit troublesome to make change since we'd have to change all of them.

.ci/AML-nightly-gpu.yml
```diff
@@ -37,20 +18,147 @@ pr:
...
   - name  : 'expname'
     value : 'nightly-smoke-test-linux-gpu'
+  - name  : 'test'
+    value : '.ci/run_pytest.py'
+  - name  : 'maxnodes'
+    value : 4
+  - name  : 'clustername'
+    value : 'cv-nightly-gpu'
+  - name  : 'vmsize'
+    value : 'STANDARD_NC6'
```
jiata: we might want to move off NC6 to at least NC6_v2. With object detection, this _might_ end up being too slow.

PatrickBue: I actually noticed that the integration tests on our Linux GPU run much faster than on AML.

.ci/AML-nightly-gpu.yml
bethz: No, many of those parameters are specific to the test.  For those that are not specific, we decided not to put them in a separate template as it would obfuscate the content and we decided clarity was more important.  Typically, once tests are running regularly, it's not the norm to require any changes.

bethz:
@jiata Perhaps we should talk as we don't seem to be looking at the same thing.  I can show you the difference as it is incorrect to say that most of the variables are being repeated 6 times.  For example, variables like test markers may look the same but the content is indeed different. There are a few selected variables that are currently the same, like vm_size, maxnodes, etc as we expect that those will be adjusted over time and will become specific to particular tests so that the VMs are properly sized for each test.  I have a spreadsheet that lists the variables in each test, what each is set to and what variables are not set in the yml (they may be defaulted in argparse).  Vars  like run_pytest will always be in the yml file as it is much easier to change it in one yml file when needed for debugging any test issues that arise. The spreadsheet also clearly shows what is different and what is not for all of the aml tests.

I'm on vacation the tomorrow and Friday. Why don't we talk when I get back and I'll share the spreadsheet which includes comments for choices for the vars/yml.
One other point, nightly combines two separate test targets that are two unique jobs that reside in one yml file. That may appear to have some duplicates but it acts as two separate jobs in devops.  I won't merge until we talk to ensure we are on the same page.

bethz:
@miguelgfierro No, it's not working from the yml file yet.  I left it in to make it easier to just turn it on when it does start working as turning it on will only require a click box from devops..

PR #294, opened 2019-08-14
Fast image retrieval notebook

PatrickBue:
### Description
Added notebook for fast image retrieval. Will add tests soon, before merging into staging.


PatrickBue:
@jiata Let me know if you want to provide feedback on this PR. Otherwise I will merge. Thx!

utils_cv/common/gpu.py
```diff
@@ -14,8 +14,30 @@ def which_processor():
...
+
 def linux_with_gpu():
     """Returns if machine is running an Linux OS and has a GPU"""
-    is_linux = platform.system().lower() == "linux"
     has_gpu = is_available()
-    return is_linux and has_gpu
+    return is_linux() and has_gpu
+
+
+def db_num_workers(non_windows_num_workers=16):
```
jiata: [minor] following typing convention `non_windows_num_workers: int = 16`

Also, do we want to just set a static 16? people online seem to say for num_worker should = 4 * num_GPU .

utils_cv/similarity/model.py
```diff
@@ -107,34 +108,40 @@ def compute_features_learner(
     Returns: DNN feature of the provided image.
 
     """
-    if dataset_type == DatasetType.Train:
+    # Note: In Fastai, for DatasetType.Train, only the output of complete minibatches is computed. Ie if one has 101 images, 
+    # and uses a minibatch size of 10, then len(feats) is 96 and not 101. For DatasetType.Valid this is not the case,
```
jiata: "...and uses a minibatch size of 10, then len(feats) is 96 and not 101." should that "10" be a 4, 8 or 16 or something that is divisible into 96

utils_cv/common/gpu.py
PatrickBue: Thanks, changed to the recommended type convection. Leaving the number at 16 since that is the default fast.ai is using

utils_cv/similarity/model.py
PatrickBue: Thanks. the 10 was a typo, changed to 16.

PR #310, opened 2019-09-04
Pabuehle anno

PatrickBue:
### Description
Added support for more XML annotation formats, e.g. when the xml file does not specify the image path or does so incorrectly.


utils_cv/detection/plot.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         rect_th: int = 4,
         rect_color: Tuple[int, int, int] = (255, 0, 0),
         text_size: int = 25,
-        text_font: str = "DejaVuSerifCondensed.ttf",
```
PatrickBue: On windows this font does not exist.

utils_cv/detection/dataset.py
```diff
@@ -155,104 +235,32 @@ def split_train_test(
 
         return train, test
 
-    def show_ims(
-        self, rows: int = 1, cols: int = 3, rand: bool = False
-    ) -> None:
+    def show_ims(self, rows: int = 1, cols: int = 3) -> None:
```
PatrickBue: Removed 'rand' since was not used anyway and not obvious how to e.g. plot first N images.

utils_cv/detection/dataset.py
```diff
@@ -96,38 +150,64 @@ def __init__(
             self.train_ds,
             batch_size=self.batch_size,
             shuffle=True,
-            num_workers=4,
+            num_workers=db_num_workers(),
```
PatrickBue: For windows setting num workers to zero is much faster due to known issue in pytorch

utils_cv/detection/dataset.py
PatrickBue: Fyi, show_ims() throws an error when setting cols = 1

utils_cv/detection/dataset.py
```diff
@@ -40,6 +40,62 @@ def get_transform(train: bool) -> List[object]:
     return Compose(transforms)
 
 
+def parse_pascal_voc_anno(
+    anno_path: str, labels: List[str] = None
+) -> Tuple[List[AnnotationBbox], Union[str, Path]]:
+    """ Extract the annotations and image path from labelling in Pascal VOC format.
+
+            Args:
```
jiata: [minor] formatting of the comment is weird here

utils_cv/detection/dataset.py
```diff
@@ -96,38 +150,64 @@ def __init__(
...
+            ]
+            self.im_paths = [
+                os.path.join(self.root / self.im_dir, s) for s in im_filenames
+            ]
+
+        # Parse all annotations
+        self.anno_paths = []
+        self.anno_bboxes = []
+        for anno_filename in anno_filenames:
+            anno_path = self.root / self.anno_dir / str(anno_filename)
```
jiata: maybe we should check here that the anno_path is valid since we are assuming that images and annotation files have the same name. Which is fair, but just in case.
`assert is_valid_file(anno_path)`

utils_cv/detection/dataset.py
```diff
@@ -96,38 +150,64 @@ def __init__(
...
-        # TODO implement:
-        # - option to bypass this setup set if users already know label names
-        # - option to do 'active' loading so that we preload all annotations
-        # during setup time (instead of lazy loading) - this will let us
-        # do things like get average box size of all the data, etc..
 
+        # If im_dir specified then get image paths and respective annotations
+        # from scanning that directory. Otherwise find all annotation files
+        # and get path to image from annotation.
+        if self.im_dir is None:
```
jiata: currently, you have to know that this code prioritizes using im_dir over a path in the annotation file. If you're just using this function, you won't know this. 

I would propose raising an error if BOTH im_dir is specified and `parse_pascal_voc_anno` returns an `im_path`

utils_cv/detection/dataset.py
```diff
@@ -96,38 +150,64 @@ def __init__(
...
+            anno_filenames = sorted(os.listdir(self.root / self.anno_dir))
+            self.im_paths = []
+        else:
```
jiata: this seems confusing to me. if im_dir is not specified, does that mean `self.anno_dir` is not specified either?

does this not work?
```python
if self.im_dir is None:
    ...
else:
    anno_filenames = sorted(os.listdir(self.root / self.anno_dir))
```

utils_cv/detection/dataset.py
```diff
@@ -53,8 +109,8 @@ def __init__(
         batch_size: int = 2,
         transforms: object = get_transform(train=True),
         train_pct: float = 0.5,
-        im_dir: str = "images",
         annotation_dir: str = "annotations",
+        im_dir: str = None,
```
jiata: shouldn't the default behavior be "images" instead of None?

utils_cv/detection/dataset.py
PatrickBue: In my case I moved .xml annotations to another directory and hence the path to the image is not correct anymore (and hence needs to be ignored when im_dir is specified). I will try to point this out more in the function description. 
And yes, anno_dir always has to be specified.

utils_cv/detection/dataset.py
```diff
@@ -96,38 +150,64 @@ def __init__(
...
             self.test_ds,
             batch_size=self.batch_size,
             shuffle=False,
-            num_workers=4,
+            num_workers=db_num_workers(),
             collate_fn=collate_fn,
         )
 
-    def _get_labels(self) -> List[str]:
+    def _read_annos(self) -> List[str]:
```
jiata: can we refactor to
```python

def _read_annos(self) -> List[str]:
        anno_filenames = sorted(os.listdir(self.root / self.anno_dir))
        self.im_paths = []

        # Parse all annotations
        self.anno_paths = []
        self.anno_bboxes = []
        for anno_filename in anno_filenames:
            anno_path = self.root / self.anno_dir / str(anno_filename)
            anno_bboxes, im_path = parse_pascal_voc_anno(anno_path)

            # TODO For now, ignore all images without a single bounding box in it, otherwise throws error during training.
            if len(anno_bboxes) == 0:
                continue

            # get im_path from anno or from im_dir
            if self.im_dir is None:
                self.im_paths.append(im_path)
            else:
                self.im_paths.append(os.path.splitext(anno_path)[0] + ".jpg"

            self.anno_paths.append(anno_path)
            self.anno_bboxes.append(anno_bboxes)

```

utils_cv/detection/dataset.py
PatrickBue: Good point.

PatrickBue:
Thanks! I am making the suggested edits and will then request another review (or shall I merge directly?)

jiata:
> Thanks! I am making the suggested edits and will then request another review (or shall I merge directly?)

feel free to merge! i'll unblock.

utils_cv/detection/dataset.py
PatrickBue: That's what I had first, but then changed to the current code so that I am not assuming an image ends with ".jpg" and rather I assume the annotation files end with ".xml"

PatrickBue:
@jiata Thanks for the review!

PR #311, opened 2019-09-05
Use pathlib only in data.py

simonzhaoms:
### Checklist:

### DIFF
--- a/utils_cv/common/data.py
+++ b/utils_cv/common/data.py
@@ -1,7 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-import os
 from pathlib import Path
 import requests
 from typing import List, Union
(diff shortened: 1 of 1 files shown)


simonzhaoms:
I find `os.path` and `pathlib.Path` being used interchangeably in `utils_cv/common/data.py`.  Not sure if appropriate to use only one of them.  There are [many other places](https://github.com/microsoft/ComputerVision/search?q=pathlib&unscoped_q=pathlib) doing in this way.

jiata:
> I find `os.path` and `pathlib.Path` being used interchangeably in `utils_cv/common/data.py`. Not sure if appropriate to use only one of them. There are [many other places](https://github.com/microsoft/ComputerVision/search?q=pathlib&unscoped_q=pathlib) doing in this way.

yeah maybe we should stick to pathlib.Path, as you do in your PR - seems more modern...

PatrickBue:
@simonzhaoms Please only merge this after running our full tests on windows and on Linux. I have seen inconsistent behavior before, that's why. 
Fyi, currently our windows tests are failing, so that's a blocker but Miguel is working on it.

PatrickBue:
Thanks for making this change. See my comment above and why we shouldn't merge this at the moment. Our devops testing pipeline is currently half broken. Miguel is working on it.

jiata:
pathlib should be OS agnostic, but good idea to make sure tests pass!

https://pbpython.com/pathlib-intro.html
> "Pathlib is an object oriented interface to the filesystem and provides a more intuitive method to interact with the filesystem in a platform agnostic and pythonic manner."

simonzhaoms:
Thank you.  Sure.  Just make sure we are aware of this.  And it is less error-prone.

PR #314, opened 2019-09-09
bug fixes in detection code / 01 notebook

jiata:
### Description
minor fixes


environment.yml
```diff
@@ -19,7 +19,7 @@ dependencies:
 - python==3.6.8
 - pytorch>=1.2.0
 - torchvision>=0.3.0
-#- pycocotools>=2.0
+- pycocotools>=2.0
```
PatrickBue: This line pycocotools will break the windows devops tests. How about we install cython here, but no pycocotools and rather ask the user to install manually?

utils_cv/detection/plot.py
```diff
@@ -26,9 +27,12 @@ def __init__(
         rect_th: int = 4,
         rect_color: Tuple[int, int, int] = (255, 0, 0),
         text_size: int = 25,
-        text_font: str = "arial.ttf",
+        text_font: str = "DejaVuSerifCondensed.ttf",
         text_color: Tuple[int, int, int] = (255, 255, 255),
     ):
+        if is_windows() == True:
```
PatrickBue: Does arial.tff not work on Linux? If it does not, how about we find a font which works on all OS rather than the if clause here?

README.md
```diff
@@ -27,6 +27,8 @@ To get started:
     git clone https://github.com/Microsoft/ComputerVision
     ```
 1. Install the conda environment, you'll find the `environment.yml` file in the root directory. To build the conda environment:
+    > If you are on Windows, remove the line `- pycocotools>-2.0` and uncomment
```
PatrickBue: Fyi, I added some installation description in the object detection README, which after this PR will need to get edited or removed.

environment.yml
jiata: without it, the tests will still fail once we add those notebook tests to the pipeline

utils_cv/detection/plot.py
jiata: i did a (very) quick search to see if there was one, but didnt come across one. i'll double check

README.md
jiata: maybe we should consider env yml generation like the reco repo in the future

environment.yml
PatrickBue: True, but at least all classification and similarity tests will run. We need to find a way to also support object detection tests.

README.md
PatrickBue: Yes. Let's discuss during our next 1:1. I know they have a script, which I guess means we'd have to have two: one for windows (.bat) and one for Linux (bash). Or maybe we can use setup.py. Not sure what the best option is

PatrickBue: Yes. One extra step, executing python scripts/generate_conda_file.py seems OK and it's easy to update out tests to work with that.

utils_cv/detection/plot.py
jiata: yeah can't seem to find.. but this post suggets that you have to deal with it separately https://stackoverflow.com/questions/24085996/how-i-can-load-a-font-file-with-pil-imagefont-truetype-without-specifying-the-ab

environment.yml
jiata: i uncomment both for now, and in the readme instructions, tell users to uncomment one of them depending on their os.

PR #317, opened 2019-09-11
detection tests

jiata:
### Description
initial work, more to come but not rdy for PR yet.

for now, minified data, and added tests for non-aml notebooks. At the very least we can make sure our 00 and 01 notebooks passes after this is merged.


jiata:
closing for now until i finish more unit tests.

PR #323, opened 2019-09-16
Updating environment.yml file in master

PatrickBue:
Merging environment.yaml file from staging into master. 

Am testing atm if with this change also the windows CPU devops test will run successfully.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,21 +17,26 @@ channels:
 - fastai
 dependencies:
 - python==3.6.8
-- pytorch==1.0.0
-- torchvision  
-- fastai==1.0.48
+- pytorch>=1.2.0
+- torchvision>=0.3.0
+- fastai==1.0.57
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -26,12 +26,16 @@ dependencies:
...
 - pip>=19.0.3
+- cython>=0.29.1
+- papermill>=0.15.0
+- black>=18.6b4
+- ipywebrtc 
+- lxml>=4.3.2
+- pre-commit>=1.14.4
+- pyyaml>=5.1.2
+- requests>=2.22.0
+- cython
```
miguelgfierro: cython is twice, apart from that, this looks good

PatrickBue: Good catch! Let's use this .yaml file for now to unblock you in master. We will update very soon anyway when merging staging into master. Will remove the second cython now in staging.

PR #326, opened 2019-09-17
Pytest suite for detection utils and notebooks

jiata:
### Description
All testing for detection utils and notebooks (Except aml ones)


PatrickBue:
Sorry for the delay in doing a CR..

tests/unit/detection/test_detection_model.py
```diff
@@ -0,0 +1,116 @@
...
+    model = get_pretrained_fasterrcnn(num_classes=classes, min_size=600, max_size=2000)
+    learner = DetectionLearner(od_detection_dataset, model=model)
+    assert type(learner) == DetectionLearner
+    assert learner.model == model
+    assert learner.model != get_pretrained_fasterrcnn(classes)
+
+
+def test_detection_learner_train_one_epoch(od_detection_learner):
+    """ Simply test that a small training loop works. """
+    od_detection_learner.fit(epochs=1)
```
PatrickBue: [minor] I have a feeling this might be too slow for CPU unless we reduce the training set to only a couple of images, but we'll see.

tests/unit/detection/test_detection_model.py
```diff
@@ -0,0 +1,116 @@
...
+@pytest.fixture(scope="session")
+def od_detection_dataset(tiny_od_data_path):
+    """ returns a basic detection dataset. """
+    return DetectionDataset(tiny_od_data_path)
+
+
+@pytest.fixture(scope="session")
+def od_detection_learner(od_detection_dataset):
+    """ returns a basic detection dataset. """
+    learner = DetectionLearner(od_detection_dataset)
```
PatrickBue: to speed things up for unit tests, could here reduce image resolution to e.g. 100 pixels (smaller causes errors) and also set the number of proposals to consider to something small.

tests/unit/detection/test_detection_notebooks.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@pytest.mark.notebooks
+def test_01_notebook_run(detection_notebooks, tiny_od_data_path):
+    notebook_path = detection_notebooks["01"]
+    pm.execute_notebook(
+        notebook_path,
+        OUTPUT_NOTEBOOK,
+        parameters=dict(
+            PM_VERSION=pm.__version__, DATA_PATH=tiny_od_data_path, EPOCHS=1
```
PatrickBue: Similar to above, this might be too slow for CPU-based unit tests. If so, could reduce image resolution to 100 pixels, reduce number of proposals, and only use a couple of training images.

PR #328, opened 2019-09-17
remove repo metrics folder

yijingchen:
### Description

Remove the repo metrics folder since this is not directly related to CVBP and we have set up internal ways to track the github statistics.

### Checklist:


yueguoguo:
**THIS DOES NOT RELATE THE CODES IN THE PR**

For the failing test [here](https://github.com/microsoft/ComputerVision/blob/dfabc3304296689e38d1c90b527d6688563ef09c/tests/integration/classification/test_integration_classification_notebooks.py#L86) is the max duration absolutely larger than 1.2 times of the min duration? From the [notebook], seems the duration values are generated from experiments by using different parameter values, so I guess they are non-deterministic? If so, we may consider using approximate testing, something like
```
assert nb_output.scraps["max_duration"].data pytest.approx(
    1.2 * nb_output.scraps["min_duration"].data,
    rel=TOL
)
```
where `TOL` is a pre-defined relative tolerance of the error, say 0.05. This is to test that using a different set of parameter values will differ in terms of run time and the difference is within an expected range. 

Reference [here](https://docs.pytest.org/en/latest/reference.html#pytest-approx)

PatrickBue:
I just loosened the threshold on the failing test from 1.2 to 1.05. These tests are indeed non-deterministic and given that we have a lot of them we are using pretty loose assert.  @yueguoguo thanks for your comments, we use approximate asserts e.g. when testing that two numbers are identical.

PatrickBue:
@yijingchen I just merged this PR. Thanks again!

yijingchen:
@PatrickBue Thank you very much for the quick action!

PR #329, opened 2019-09-18
Fix bug in reading VOC files

PatrickBue:
Fixed a bug I introduced which is triggered when some of the images do not have any annotated bounding boxes in them.


PatrickBue:
@jiata No it does not break or affect the 01 notebook. I now get good accuracies for all except for one of my internal datasets (that dataset for some reason the ground truth isn't read correctly)

PR #331, opened 2019-09-24
Read floating point coordinate from text instead of integer

simonzhaoms:
### Description

Changed the parsed coordinates from integer to float.

Some PASCAL VOC annotations (reported by Mirco) give coordinates of floating point which fails the conversion from text to integer by using `int(str)` in the function `parse_pascal_voc_anno()`  in `utils_cv/detection/dataset.py`.  And in `PIL.ImageDraw`, floating point coordinates works fine, so using floating point will include more use cases.

### Checklist:

### DIFF
--- a/utils_cv/detection/dataset.py
+++ b/utils_cv/detection/dataset.py
@@ -73,10 +73,10 @@ def parse_pascal_voc_anno(
     for obj in objs:
         label = obj.find("name").text
         bnd_box = obj.find("bndbox")
-        left = int(bnd_box[0].text)
-        top = int(bnd_box[1].text)
-        right = int(bnd_box[2].text)
-        bottom = int(bnd_box[3].text)
+        left = float(bnd_box[0].text)
+        top = float(bnd_box[1].text)
+        right = float(bnd_box[2].text)
+        bottom = float(bnd_box[3].text)
 
         # Set mapping of label name to label index
         if labels is None:


simonzhaoms:
> I believe these co-ordinates should indeed be integers since these are absolute pixel coordinates and not relative coordinates. Or is there a good reason why these should be floats? @jiata

Yes, indeed.  But the predicted coordinates from object detection models as well as some from PASCAL VOC annotations are actually floats.  I thought to use `int(float(str))` instead before I realized `PIL.ImageDraw` use `double` underneath:

https://github.com/python-pillow/Pillow/blob/349e220b0a7e9396843e039c5f481d545d28fa91/src/_imaging.c#L3002-L3037

So for drawing with `PIL`, floats are fine.  But when using the returned `AnnotationBox`, it needs to be considered to convert floats to ints.

PatrickBue:
Sorry I am still confused why a point co-ordinate should be float and not integer, except perhaps for the highly special case where we need sub-pixel accuracy. We also treat width and height as integers. In the past, I had the problem actually that my co-ordinates were floats and had to convert them to integer every time I wanted to e.g. draw a rectangle using OpenCV. 
How about we keep this PR dormant for now, and revisit if we have a stronger point why this should be float and not int?

simonzhaoms:
> Sorry I am still confused why a point co-ordinate should be float and not integer, except perhaps for the highly special case where we need sub-pixel accuracy. We also treat width and height as integers. In the past, I had the problem actually that my co-ordinates were floats and had to convert them to integer every time I wanted to e.g. draw a rectangle using OpenCV.
> How about we keep this PR dormant for now, and revisit if we have a stronger point why this should be float and not int?

Sure.  Close this PR or just leave it open?

PatrickBue:
> Sorry I am still confused why a point co-ordinate should be float and not integer, except perhaps for the highly special case where we need sub-pixel accuracy. We also treat width and height as integers. In the past, I had the problem actually that my co-ordinates were floats and had to convert them to integer every time I wanted to e.g. draw a rectangle using OpenCV.
> How about we keep this PR dormant for now, and revisit if we have a stronger point why this should be float and not int?
> 
> Sure.  Close this PR or just leave it open?

I would suggest up to you. Feel free to leave it open and close later, or close now.

simonzhaoms:
OK.  Just leave it open :smile:

PR #353, opened 2019-09-27
fixed annotation loading for VOC dataset

PatrickBue:


PatrickBue:
> looks good, as long as tests pass (which its looking like it does!)

Yes.. I will wait for that, now that we have tests again. Yay. Fyi, we should probably also trigger our VM linux + windows CPU tests (currently only the GPU tests are triggered above)

jiata:
> > looks good, as long as tests pass (which its looking like it does!)
> 
> Yes.. I will wait for that, now that we have tests again. Yay. Fyi, we should probably also trigger our VM linux + windows CPU tests (currently only the GPU tests are triggered above)

ok done.

Issue #350, opened 2019-09-26
Installation update


jiata:
### Description
Update installation instructions to point out that for image classification, Linux is up to 10x faster in training than Linux (and if not setting num_workers=0 then windows even two times slower) 

For object detection, windows is only ~20% slower in training but about same speed for inference

PR #358, opened 2019-09-30
README updates

jiata:
### Description
restructure readme:
- move getting started to a new `setup.md` page
- move notebook specific things into the readme of the `scenario` folder
- streamline text/ remove redundancy

#350


jiata:
for those who are reviewing, would recommend reviewing directly in the branch: https://github.com/microsoft/ComputerVision/tree/jiata-minor-fixes

environment.yml
```diff
@@ -34,6 +34,7 @@ dependencies:
 - pre-commit>=1.14.4
 - pyyaml>=5.1.2
 - requests>=2.22.0
+- cython>=0.29.1
```
PatrickBue: This is a duplicate. See a bit further above in this file.

SETUP.md
```diff
@@ -0,0 +1,121 @@
...
+Specifically, we'll want one with a powerful GPU. The NVIDIA's Tesla V100 is a
+good choice that can be found in most Azure regions.
+
+The easiest way to get started is to use the [Azure Data Science Virtual Machine (DSVM)](https://azure.microsoft.com/en-us/services/virtual-machines/data-science-virtual-machines/). This VM will come installed with all the system requirements that are needed to run the notebooks in this repository. If you choose this option, you can skip the [System Requirements](#system-requirements) step in this guide as those requirements come pre-installed on the DSVM.
+
+Here are some ways you can create the DSVM:
+
+__Virtual Machine Builder__
+
+One easy way to create your DSVM is to use the [VM Builder](../contrib/vm_builder) tool located inside of the 'contrib' folder in the root directory of the repo. Simply run `python contrib/vm_builder/vm_builder.py` at the root level of the repo and this tool will preconfigure your virtual machine with the appropriate settings for working with this repository.
```
PatrickBue: This section should probably be moved until after the "provision a data science VM.." section and pointed out that our recommendation is to use the portal. That's partly since the portal is easy to use and I think good for beginners to know what exactly they are creating and how to delete it. And partly since the tool could stop working (since we don't test it).

scenarios/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; <img align="center" src="./media/intro_is_vis.jpg" height="150" alt="Image similarity visualization"/>
+
+- **Image Segmentation** Given an input image, assign a label to every pixel (e.g., background, bottle, hand, sky, etc.). In practice, this problem is less common in industry, in large part due to time required to label the ground truth segmentation required in order to train a solution.
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; <img align="center" src="./media/intro_iseg_vis.jpg" height="150" alt="Image segmentation visualization"/>
+
+# Data/Telemetry
+
+The following notebooks collect browser usage data and sends it to Microsoft to help improve our product and services:
```
PatrickBue: Nice. This looks much better than before.

SETUP.md
```diff
@@ -0,0 +1,121 @@
...
+You can alternatively use the Azure command line (CLI) as well. Follow
+[this](https://docs.microsoft.com/en-us/cli/azure/azure-cli-vm-tutorial?view=azure-cli-latest)
+link to learn more about the Azure CLI and how it can be used to provision
+resources.
+
+## System Requirement
+
+__Requirements__
+
+* A machine running Linux >= 16.04 LTS or Windows
```
PatrickBue: Need to point out here (and maybe also at other places) that especially for image classification we recommend Linux. That's because PyTorch runs much slower on Windows, for image classification, training on Linux can be up to 10x faster. For object detection, the difference is much smaller, and around 20-30%. See our OneNote page for more detailed info.

jiata: that is already noted a few paragraphs below

Issue #336, opened 2019-09-26
[BUG] image transformations are not applied correctly


jiata:
### Description
Bug with train/test split, looks like right now neither training nor testing gets horizontally flipped (better than both getting flipped) 

"had a look at how many transformations there are in the train and test set. Both only contained a single transformation, but the training set should be two"

PR #360, opened 2019-10-02
transforms fix

jiata:
### Description
#336 

Fix on transforms


PatrickBue:
@jiata fyi this PR is to master

utils_cv/detection/dataset.py
```diff
@@ -34,7 +35,7 @@ def get_transform(train: bool) -> List[object]:
     """
     transforms = []
     transforms.append(ToTensor())
-    if train:
+    if train is True:
```
PatrickBue: [minor] do we want the "is True" here?

utils_cv/detection/dataset.py
```diff
@@ -34,7 +35,7 @@ def get_transform(train: bool) -> List[object]:
...
         transforms.append(RandomHorizontalFlip(0.5))
```
PatrickBue: Could you add this transformation here always for train:
torchvision.transforms.ColorJitter(brightness=0.2, contrast=0.2, saturation=0.4, hue=0.05)

utils_cv/detection/dataset.py
```diff
@@ -231,17 +231,14 @@ def split_train_test(
         Return
             A training and testing dataset in that order
         """
-        # TODO Is it possible to make these lines in split_train_test() a bit
-        # more intuitive?
-
         test_num = math.floor(len(self) * (1 - train_pct))
         indices = torch.randperm(len(self)).tolist()
 
         self.transforms = get_transform(train=True)
```
PatrickBue: I find this syntax quite confusing. Could rewrite to:
        train = Subset(self, indices[test_num:])
        train.dataset.transforms = get_transform(train=True)

        test = Subset(self, indices[: test_num + 1])
        test.dataset.transforms = get_transform(train=False)

PatrickBue:
@jiata I am leaving some minor comments now. Have a look if this PR is merging more into staging that wanted (e.g. changes to README).

utils_cv/detection/dataset.py
jiata: oh yeah i was trying many things to test, forgot to remove this.

jiata:
> @jiata I am leaving some minor comments now. Have a look if this PR is merging more into staging that wanted (e.g. changes to README).

that was because i accidentally set base to master earlier

utils_cv/detection/dataset.py
jiata: done

utils_cv/detection/dataset.py
jiata: done

utils_cv/detection/dataset.py
jiata: done

utils_cv/detection/dataset.py
jiata: done

utils_cv/detection/dataset.py
PatrickBue: thanks!

utils_cv/detection/dataset.py
```diff
@@ -231,17 +235,14 @@ def split_train_test(
...
         """
-        # TODO Is it possible to make these lines in split_train_test() a bit
-        # more intuitive?
-
         test_num = math.floor(len(self) * (1 - train_pct))
         indices = torch.randperm(len(self)).tolist()
 
-        self.transforms = get_transform(train=True)
-        train = Subset(self, indices[test_num:])
+        train = copy.deepcopy(Subset(self, indices[test_num:]))
```
PatrickBue: IS the copy.deepcopy() required? I my code it seems to be working without

jiata: oh really? i thought i had tested it without and it didn't work... let me verify

jiata: yeah, it does __not__ work without the deepcopy

PatrickBue: Strange, let's use the copy then, though for me it seems to be working withouth. By the way, this class takes as input "self.transforms" and looks like this input is not used anywhere @jiata

jiata: thats a good point, actually its misleading even. let me do a refactor...

utils_cv/detection/dataset.py
```diff
@@ -107,7 +112,8 @@ def __init__(
         self,
         root: Union[str, Path],
         batch_size: int = 2,
-        transforms: object = get_transform(train=True),
+        train_transforms: object = get_transform(train=True),
```
PatrickBue: nice!

utils_cv/detection/dataset.py
```diff
@@ -36,6 +38,9 @@ def get_transform(train: bool) -> List[object]:
     transforms.append(ToTensor())
     if train:
         transforms.append(RandomHorizontalFlip(0.5))
+        transforms.append(
+            ColorJitter(brightness=0.2, contrast=0.2, saturation=0.4, hue=0.05)
+        )
         # TODO we can add more 'default' transformations here
```
PatrickBue: [minor] can remove Todo here

simonzhaoms:
How about the below based on the commit on staging on Sep 28, 2019 (2e3deb4): 

```python
class DetectionDataset(Dataset):
    def __init__(
        self,
        root: Union[str, Path],
        batch_size: int = 2,
        transforms: Union[Trans, Tuple[Trans, Trans]] = (
                get_transform(train=True),
                get_transform(train=False)
        ),
        train_pct: float = 0.5,
        im_dir: str = "images",
    ):
        ...
        ...
        if transforms and len(transforms) == 1:
            self.transforms = (transforms, ) * 2
        self.transforms = transforms
        ...
        ...

    def split_train_test(
        self, train_pct: float = 0.8
    ) -> Tuple[Dataset, Dataset]:
        ...
        ...
        self.is_test = np.zeros((len(self),), dtype=np.bool)
        self.is_test[test_idx] = True
        ...
        ...

    def _get_transforms(self, idx):
        """ Return the corresponding transforms for training and testing data. """
        return self.transforms[self.is_test[idx]]

    def __getitem__(self, idx):
        ...
        ...
        if self.transforms:
            im, target = self._get_transforms(idx)(im, target)
        ...
        ...
```

jiata:
seems like a bit of a complicated way to determine which transforms are for test vs train when we can just use to params to store them. 

Is there an advantage to have the 'is_test' over the subset method?

simonzhaoms:
> seems like a bit of a complicated way to determine which transforms are for test vs train when we can just use to params to store them.
> 
> Is there an advantage to have the 'is_test' over the subset method?

seems if 
```python
test = copy.deepcopy(Subset(self, indices[: test_num + 1]))
```
then the whole original dataset (`self`) will have two copies.  If I guess correctly, may it break the intention of using `self` to manipulate training and test datasets simultaneously in `DetectionDataset` without duplicating the data itself and data preprocessing?

jiata:
> > seems like a bit of a complicated way to determine which transforms are for test vs train when we can just use to params to store them.
> > Is there an advantage to have the 'is_test' over the subset method?
> 
> seems if
> 
> ```python
> test = copy.deepcopy(Subset(self, indices[: test_num + 1]))
> ```
> 
> then the whole original dataset (`self`) will have two copies. If I guess correctly, may it break the intention of using `self` to manipulate training and test datasets simultaneously in `DetectionDataset` without duplicating the data itself and data preprocessing?

Thats actually a really good point, thanks for bringing it up.

jiata:
so i looked at the torchvision sample code in their docs and their code also replicates two versions of the dataset. 

They do something like this:
```
dataset = Dataset(..., get_transforms(train=True))
dataset_test = Dataset(..., get_transforms(train=False))

dataset = Subset(...)
dataset_test = Subset(...)
...
```

In this case, they also have two copies of the dataset object (since `Subset` keeps the reference).

I'll merge the PR in for now.

utils_cv/detection/dataset.py
```diff
@@ -256,6 +289,43 @@ def show_ims(self, rows: int = 1, cols: int = 3) -> None:
         """
         plot_grid(display_bboxes, self._get_random_anno, rows=rows, cols=cols)
 
+    def show_im_transformations(
```
PatrickBue: Nice! We could demonstrate this in on of our notebooks, with an explanation when not to use e.g. horizontal flipping.

PR #364, opened 2019-10-08
Correct unzipped file dest folder to unzipped_dir

simonzhaoms:
### Description

The function `unzip_url()` is used to download a zip file to `fpath / zip file` and unzip it to `dest / zip file without extension`, so the correct parameter for `ZipFile.extratcall()` needs to be `dest / zip file without extension` which is `unzipped_dir` in the code instead of `fpath`.

### Checklist:

### DIFF
--- a/scenarios/classification/01_training_introduction.ipynb
+++ b/scenarios/classification/01_training_introduction.ipynb
@@ -104,7 +104,7 @@
    },
    "outputs": [],
    "source": [
-    "DATA_PATH     = unzip_url(Urls.fridge_objects_path, exist_ok=True)\n",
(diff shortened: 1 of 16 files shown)


jiata:
can you update the tests too?

simonzhaoms:
> we need to update the tests too

OK, I will update related code

simonzhaoms:
Updated

simonzhaoms:
After discussion, we decide to keep the function as it is.

PatrickBue:
@simonzhaoms Thanks for talking to JS and deciding on what the best way forward is!!

PR #367, opened 2019-10-12
Adding hard negative sampling notebook

PatrickBue:


PatrickBue:
@jiata Fyi, here is the hard-neg sampling notebook. It's working, but will make a few more edits on Monday (mainly to the text) and then send out a code review invite.

utils_cv/detection/dataset.py
```diff
@@ -93,7 +93,7 @@ def parse_pascal_voc_anno(
 
     # get image path from annotation. Note that the path field might not be set.
     anno_dir = os.path.dirname(anno_path)
-    if root.find("path"):
+    if root.find("path") is not None:
```
PatrickBue: This is to remove a warning which python otherwise shows

utils_cv/detection/dataset.py
```diff
@@ -146,6 +146,7 @@ def __init__(
         train_pct: float = 0.5,
         anno_dir: str = "annotations",
         im_dir: str = "images",
+        require_annotation_files = True
```
PatrickBue: Let me know if you can think of a better name for this variable

utils_cv/detection/dataset.py
```diff
@@ -218,20 +208,33 @@ def _read_annos(self) -> List[str]:
...
+                im_path = im_paths[anno_idx]
+
+            # Torchvision needs at least one ground truth bounding box per image. Hence for images without a single
+            # annotated object, adding a tiny bounding box with "background" label 0.
             if len(anno_bboxes) == 0:
-                continue
+                anno_bboxes = [
+                    AnnotationBbox.from_array(
+                        [1, 1, 5, 5],
+                        label_name=None,
```
PatrickBue: Note the special case here: I am setting the label_name to None since this label does not correspond to any of the objects-of-interest. Alternative would be to set it to "__background__", but that would also require some code edits so that nothing breaks

utils_cv/detection/dataset.py
```diff
@@ -273,11 +280,55 @@ def split_train_test(
         train = copy.deepcopy(Subset(self, indices[test_num:]))
         train.dataset.transforms = self.train_transforms
 
-        test = copy.deepcopy(Subset(self, indices[: test_num + 1]))
+        test = copy.deepcopy(Subset(self, indices[: test_num]))
```
PatrickBue: One image ended up in the training AND in the test set, hence this fix.

utils_cv/detection/dataset.py
```diff
@@ -218,20 +208,33 @@ def _read_annos(self) -> List[str]:
...
+            # Parse annotation file if present
+            if os.path.exists(anno_path):
+                anno_bboxes, im_path = parse_pascal_voc_anno(anno_path)
+            else:
+                if self.require_annotation_files:
+                    raise FileNotFoundError(anno_path)
+                anno_bboxes = [] 
+                im_path = im_paths[anno_idx]
+
+            # Torchvision needs at least one ground truth bounding box per image. Hence for images without a single
```
jiata: is this the right way to handle this case - adding a tiny bbox with label 0?

utils_cv/detection/dataset.py
```diff
@@ -163,6 +164,8 @@ def __init__(
             train_pct: the ratio of training to testing data
             annotation_dir: the name of the annotation subfolder under the root directory
             im_dir: the name of the image subfolder under the root directory. If set to 'None' then infers image location from annotation .xml files
+            require_annotation_files: is true (default) then will throw an error if no anntation can be found for a given image. Otherwise use image
```
jiata: maybe `allow_negatives`?

jiata: default should arguable be false

utils_cv/detection/dataset.py
```diff
@@ -273,11 +280,55 @@ def split_train_test(
...
+            if target.lower() == "train":
+                self.train_ds.dataset.im_paths.append(im_path)
+                self.train_ds.dataset.anno_bboxes.append(anno_bbox)
+                self.train_ds.indices.append(len(self.im_paths)-1)
+            elif target.lower() == "test":
+                self.test_ds.dataset.im_paths.append(im_path)
+                self.test_ds.dataset.anno_bboxes.append(anno_bbox)
+                self.test_ds.indices.append(len(self.im_paths)-1)
+            else:
+                raise Exception("Target " + str(target) + " unknown.")
```
jiata: f-string format

utils_cv/detection/dataset.py
```diff
@@ -273,11 +280,55 @@ def split_train_test(
...
+            collate_fn=collate_fn,
+        )
+
+    def add_images(self, im_paths: List[str], anno_bboxes: List[AnnotationBbox], target: str = "train"):
+        """ Add new images to either the training or test set. 
+
+        Args:
+            im_paths: path to the images.
+            anno_bboxes: ground truth boxes for each image.
+            target: specify if images are to be added to the training or test set. Valid options: "train" or "test".
```
jiata: we can add raise to docstring
ex:
```
    Raises:
        Exception if `target` variable is neither 'train' nor 'test'
```

utils_cv/detection/plot.py
```diff
@@ -56,7 +56,12 @@ def plot_boxes(
     """
     if len(bboxes) > 0:
         draw = ImageDraw.Draw(im)
+        font = get_font(size=plot_settings.text_size)
+
         for bbox in bboxes:
+            # do not draw background bounding boxes
+            if bbox.label_idx == 0:
```
jiata: this is why i'm worried about the idea of adding the little box int he corner. If people wanna write their own function for plotting, they'd have to know this weird edge-case.

utils_cv/detection/dataset.py
```diff
@@ -273,11 +280,55 @@ def split_train_test(
...
+
+        self.test_dl = DataLoader(
+            self.test_ds,
+            batch_size=self.batch_size,
+            shuffle=False,
+            num_workers=db_num_workers(),
+            collate_fn=collate_fn,
+        )
+
+    def add_images(self, im_paths: List[str], anno_bboxes: List[AnnotationBbox], target: str = "train"):
```
jiata: why does this function take in anno_bboxes? in the notebook, wont all the gt anno_bboxes for these just be the tiny rectangle in the top let?

maybe this function should be called `add_negative_images` or something less generic.

utils_cv/detection/dataset.py
PatrickBue: Agree this is a hack. I read somewhere that this is a known issue (cannot find it on tensorflow anymore, thing it was in one of the "issues"). I don't know a way around this, esp without modifying the code in "references".

utils_cv/detection/dataset.py
PatrickBue: This function can be used to add more negative or positive images. Hence the anno_bboxes as required input.

utils_cv/detection/dataset.py
PatrickBue: Thanks, changed

utils_cv/detection/plot.py
PatrickBue: Fully agree. I don't like it either.

PatrickBue: Just created an issue to keep track of this. But main problem I see is that edits would at the very least be required in the "references" folder which is not good either.
https://github.com/microsoft/ComputerVision/issues/372

PR #371, opened 2019-10-15
Docker support for CVBP

yueguoguo:
### Description
Dockerfile for CVBP 

### Related Issues
N/A

### Checklist:


docker/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+ENV PATH="${HOME}/conda/bin:${PATH}"
+ENV PATH="${HOME}/conda/envs/cv/bin:${PATH}"
+
+
+# Final Stage
+FROM $ENV AS final
+
+# Install Conda dependencies
+RUN conda env update -f computervision/environment.yml && \
```
gramhagen: This updates the base env, so there won't be a CV env created. You should remove the previous lines adding the CV env to the path, or create a new env and remove the base env paths

docker/README.md
```diff
@@ -0,0 +1,60 @@
+Docker Support
```
PatrickBue: This is great, thanks for adding docker support! We aim to have tests with every PR that adds functionality. Do you think the 'docker build' command is something that should/could be added as a devops tests? How long does it take to run?

jiata:
Can you retrigger the PR - i should've fixed the windows machine

docker/Dockerfile
yueguoguo: Slightly different from reco, the yaml file that CV is using has the name specified so the `conda env` creates a cv env by default.

gramhagen: But this is running conda update. That will still create a new env?

docker/README.md
yueguoguo: It takes 16min to finish the build of a Docker image. We can probably have a specific build pipeline that runs testing on the Docker build. It may not be triggered by PR as I don't expect the Dockerfile will be updated with high frequency. What do you think?

yueguoguo:
> Can you retrigger the PR - i should've fixed the windows machine
Manually rerun the build and it seems not working... Not sure why the module `fastai` cannot be found.

docker/Dockerfile
yueguoguo: Ah I see if not exist `conda update` will create a new one but yea it is more clear to use `conda create` in this context. 

Updated the PR.

docker/README.md
PatrickBue: @jiata Adding JS to this thread. 16min isn't too bad, and acceptable for our over-night devops pipeline. See the tests under tests\smoke\test_azureml_notebooks.py, they also take up to 30min each. 
But agree, if the test only verifies that the docker build commend worked, then this test might not be needed. Ideally we'd also want to test that CVBP runs successfully inside the docker container, but not sure that's doable?

PatrickBue: @yueguoguo are the other repos doing or planning on doing any sort of testing around the docker containers? Also, unrelated, are the othee repos hosting build docker containers for download?

docker/Dockerfile
gramhagen: Ok, I didn't realize that, but I agree making it more explicit is good. In this case the two lines that add the base probably should be removed

```ENV PATH="${HOME}/conda/bin:${PATH}"```

this might lead to confusion if something breaks in the cv env it would fall back to the base which wouldn't have the right packages.

jiata:
@yueguoguo how do we know that the build worked inside docker? is there some way to test? or is it manual for now?

docker/Dockerfile
yueguoguo: Yea we can remove this.

docker/README.md
yueguoguo: @PatrickBue reco does not have testing pipeline for the Docker file, and nlp repo does not have Docker support at the moment. It makes sense for having testing for the Docker things, but not sure if it is sensible to run all the testing. 

@jiata to run test within a Docker container, see [here](https://github.com/microsoft/recommenders/tree/master/docker#running-tests-with-docker).

yueguoguo:
@PatrickBue please let me know if there is anything before the PR can be merged. Thanks!

PR #389, opened 2019-11-06
Add Mask R-CNN

simonzhaoms:
### Description

Add Mask R-CNN notebook.  

This PR was created for figuring out whether a PR from a fork won't trigger all testing pipelines.  Since this PR triggered all testing pipeline, it now replaces https://github.com/microsoft/ComputerVision/pull/363 to be the main PR for Mask R-CNN notebook.

### Checklist:

- [X] I have added tests covering my contributions.
- [X] I have updated the documentation accordingly.
- [X] This PR is being made to `staging` and not `master`


tests/unit/detection/test_detection_model.py
```diff
@@ -69,71 +78,133 @@ def test_detection_learner_init_model(od_detection_dataset):
 
 
 @pytest.mark.gpu
-def test_detection_learner_train_one_epoch(od_detection_learner):
```
jiata: would it make sense to split this up between detection and mask? i think it might be cleaner (but more tests)

tests/unit/detection/test_detection_model.py
```diff
@@ -69,71 +78,133 @@ def test_detection_learner_init_model(od_detection_dataset):
...
+    od_detection_mask_learner
+):
     """ Simply test that a small training loop works. """
     od_detection_learner.fit(epochs=1)
+    # test mask learner
+    od_detection_mask_learner.fit(epochs=1)
 
 
 @pytest.mark.gpu
-def test_detection_learner_plot_precision_loss_curves(od_detection_learner):
```
jiata: maybe split this up too. and all the ones below as well.

tests/unit/detection/test_detection_model.py
```diff
@@ -69,71 +78,133 @@ def test_detection_learner_init_model(od_detection_dataset):
...
     assert type(bboxes) == list
+    # test mask learner
+    bboxes, masks = od_detection_mask_learner.predict(od_cup_path, threshold=0.1)
+    assert type(bboxes) == list
+    assert type(masks) == np.ndarray
+    assert len(bboxes) == len(masks)
 
 
 @pytest.mark.gpu
 def test_detection_learner_predict_threshold(
```
jiata: this one definitely feels like it *should* be split up between detection and mask

tests/conftest.py
```diff
@@ -345,6 +349,15 @@ def od_cup_path(tmp_session) -> str:
     return im_path
 
 
+@pytest.fixture(scope="session")
+def od_cup_mask_path(tmp_session) -> str:
+    """ Returns the path to the downloaded cup image. """
+    im_url = "https://github.com/simonzhaoms/testdata/raw/master/unittest/segmentation-masks/cvbp_cup.png"
```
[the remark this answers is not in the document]
simonzhaoms: Sure.  I've just never used blob before, so I don't know how to do it.  Any guidance or conventions?  I can put this and odFridgeObjects-mask dataset into blob as well.

tests/unit/detection/test_detection_model.py
simonzhaoms: OK

tests/unit/detection/test_detection_model.py
simonzhaoms: Done

tests/conftest.py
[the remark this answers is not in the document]
jiata: will msg you offline about it

utils_cv/detection/model.py
```diff
@@ -285,63 +380,138 @@ def plot_precision_loss_curves(
...
+            ax1.set_title("Loss and Average Precision over epochs")
+            ax1.set_xlabel("epochs")
+            ax1.set_ylabel("loss", color="g")
+            ax1.plot(self.losses, "g-")
+
+            ax2 = ax1.twinx()
+            ax2.set_ylabel("average precision for ".format(k), color="b")
+            ax2.plot(v, "b-")
+
+    def evaluate(self, dl: DataLoader = None) -> Union[CocoEvaluator, None]:
```
jiata: should return a CocoEvaluator or fail. should *not* return None

utils_cv/detection/model.py
```diff
@@ -285,63 +380,138 @@ def plot_precision_loss_curves(
...
+    ) -> torch.Tensor:
+        """ Convert the image to the format required by the model. """
+        transform = transforms.Compose([transforms.ToTensor()])
+        im = transform(im)
+        if self.device:
+            im = im.to(self.device)
+        return im
+
+    @classmethod
+    def _apply_threshold(
```
jiata: there are now two `_apply_threshold` methods. Can you delete the old one?

Also, out of curiousity, why make this a cls method? We're never going to call `DetectionLearner._apply_threshold` are we?

utils_cv/detection/model.py
```diff
@@ -285,63 +380,138 @@ def plot_precision_loss_curves(
...
+        pred = {k: v.detach().cpu().numpy() for k, v in pred.items()}
+        # apply score threshold
+        if threshold:
+            selected = pred['scores'] > threshold
+            pred = {k: v[selected] for k, v in pred.items()}
+        if "masks" in pred and mask_threshold:
+            pred["masks"] = pred["masks"] > mask_threshold
+        return pred
+
+    def _get_det_bboxes(
```
jiata: same with this. now there are two `_get_det_bboxes` methods

utils_cv/detection/model.py
```diff
@@ -285,63 +380,138 @@ def plot_precision_loss_curves(
...
+        """ Gets the bounding boxes and labels from the prediction object. """
+        return DetectionBbox.from_arrays(
+            pred['boxes'].tolist(),
+            score=pred['scores'].tolist(),
+            label_idx=pred['labels'].tolist(),
+            label_name=np.array(self.labels)[pred['labels'] - 1].tolist(),
+            im_path=im_path,
+        )
+
+    def _process_pred_results(
```
jiata: this function seems extraneous. Instead of `_get_det_bboxes`, maybe we can expand the function to `_get_det_bboxes_and_mask`, and then we won't need this function anymore. i think its also semantically a bit clearer

utils_cv/detection/model.py
```diff
@@ -285,63 +380,138 @@ def plot_precision_loss_curves(
...
+
+        res = self._get_det_bboxes(pred, im_path)
+        if "masks" in pred:
+            res = (res, pred["masks"].squeeze(1))
+        return res
+
+    @classmethod
+    def _pack_pred_results(cls, res: List, infos: Dict) -> List[Dict]:
+        """ Pack up bounding boxes and mask into list of dict. """
+        if not isinstance(res[0], tuple):
```
jiata: is there a better way to pack this? the data structure is such that we have to remember that if we see a tuple, it means mask and bboxes. Maybe this should be a dict of something.

utils_cv/detection/model.py
```diff
@@ -285,63 +380,138 @@ def plot_precision_loss_curves(
...
+        im_path: Union[str, Path]
+    ) -> Union[List[Type[_Bbox]], Tuple]:
+        """ Extract bounding boxes and masks from the prediction object. """
+
+        res = self._get_det_bboxes(pred, im_path)
+        if "masks" in pred:
+            res = (res, pred["masks"].squeeze(1))
+        return res
+
+    @classmethod
```
jiata: should this be a cls method? whats the rationale

utils_cv/detection/references/coco_eval.py
```diff
@@ -114,7 +114,9 @@ def prepare_for_coco_segmentation(self, predictions):
 
             rles = [
                 mask_util.encode(
-                    np.array(mask[0, :, :, np.newaxis], order="F")
+                    # Change according to the issue related to mask:
+                    #     https://github.com/pytorch/vision/issues/1355#issuecomment-544951911
+                    np.array(mask[0, :, :, np.newaxis], dtype=np.uint8, order="F")
```
jiata: Looks good, i'll make a separate issue to do an update on the references.

tests/conftest.py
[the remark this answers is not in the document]
simonzhaoms: Thanks.  Done.  BTW, I notice there are watermark dataset, so what is for? does mask dataset need watermark?

utils_cv/detection/data.py
```diff
@@ -21,13 +21,17 @@ class Urls:
         base, "odFridgeObjectsWatermarkTiny.zip"
     )
 
+    # mask datasets
+    fridge_objects_mask_path = 'https://github.com/simonzhaoms/testdata/raw/master/odFridgeObjects-mask.zip'
```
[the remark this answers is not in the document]
simonzhaoms: Thanks.  Done.  BTW, I notice there are watermark dataset, so what is for? does mask dataset need watermark?

PatrickBue:
@simonzhaoms Great to see this PR progressing well!! I will do a code review also on Monday, but don't expect much additional comments to what JS already pointed out.

simonzhaoms:
<p>@PatrickBue Please comment on the new PR https://github.com/microsoft/ComputerVision/pull/389</p><p> </p>

utils_cv/common/plot.py
```diff
@@ -45,25 +45,34 @@ def line_graph(
 
 
 def show_ims(
-    im_paths: Union[str, List[str]],
+    im_paths: Union[str, List[str], np.ndarray, List[np.ndarray]],
```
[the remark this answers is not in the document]
simonzhaoms: Done

utils_cv/common/plot.py
```diff
@@ -45,25 +45,34 @@ def line_graph(
...
     Args:
-        im_paths (str or List[str]): Image filepaths
+        im_paths (str or List[str] or numpy.ndarray or List[numpy.ndarray]): Image filepaths
         labels (str or List[str]): Image labels. If None, show image file name.
         size (int): MatplotLib plot size.
         rows (int): rows of the images
     """
-    if isinstance(im_paths, (str, Path)):
+    if im_paths is None or len(im_paths) == 0:
+        return
```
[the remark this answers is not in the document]
simonzhaoms: Done

utils_cv/detection/dataset.py
```diff
@@ -141,12 +145,14 @@ def __init__(
...
-        train_transforms: object = get_transform(train=True),
-        test_transforms: object = get_transform(train=False),
+        train_transforms: Trans = get_transform(train=True),
+        test_transforms: Trans = get_transform(train=False),
         train_pct: float = 0.5,
         anno_dir: str = "annotations",
         im_dir: str = "images",
-        allow_negatives: bool = False,
+        mask_dir: str = None,
+        seed: int = None,
```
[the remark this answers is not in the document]
simonzhaoms: Also in `DetectionDataset.show_ims()` of `utils_cv/detection/dataset.py`, I assume we use `self.seed` if no `seed` provided.  Not sure whether it is good or not.

utils_cv/detection/model.py
```diff
@@ -285,63 +380,138 @@ def plot_precision_loss_curves(
...
+            ax1.plot(self.losses, "g-")
+
+            ax2 = ax1.twinx()
+            ax2.set_ylabel("average precision for ".format(k), color="b")
+            ax2.plot(v, "b-")
+
+    def evaluate(self, dl: DataLoader = None) -> Union[CocoEvaluator, None]:
         """ eval code on validation/test set and saves the evaluation results in self.results. """
         if dl is None:
+            if not self.dataset:
```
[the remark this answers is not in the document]
simonzhaoms: Done

utils_cv/detection/model.py
simonzhaoms: Done

utils_cv/detection/references/coco_eval.py
simonzhaoms: Today I noticed in [the newest version of references](https://github.com/pytorch/vision/blob/v0.5.0/references/detection/coco_eval.py) they fixed the bug.

utils_cv/detection/model.py
simonzhaoms: I have no idea.  My thinking is since no other places need to use this method and it is only used in `DetectionLearner`, making it a method of `DetectionLearner` would be easier to organize. That is also why a underscore is prefixed.  What do you think?

utils_cv/detection/model.py
simonzhaoms: Removed the old one.

utils_cv/detection/model.py
simonzhaoms: The same rationale as `_apply_threshold()`.  But actually I have no idea whether it is better to make it a cls method or a separate funciton.

utils_cv/detection/model.py
simonzhaoms: :+1:  Good point!  Thanks!  Done

utils_cv/common/plot.py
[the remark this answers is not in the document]
simonzhaoms: Done

utils_cv/common/plot.py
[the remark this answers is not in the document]
simonzhaoms: Done

utils_cv/detection/model.py
simonzhaoms: Merged with `_process_pred_results()` into `_get_det_bboxes_and_mask()`

utils_cv/detection/model.py
simonzhaoms: :+1:  Good point!  Thanks!  Done

utils_cv/detection/model.py
simonzhaoms: Good point! Done

utils_cv/detection/references/coco_eval.py
simonzhaoms: It was fixed in the latest version: https://github.com/pytorch/vision/blob/v0.5.0/references/detection/coco_eval.py#L112

utils_cv/detection/data.py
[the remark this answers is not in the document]
jiata: i used it for tests, but not critical

PatrickBue:
@simonzhaoms @gramhagen The notebook looks great and exactly what I had in mind. The user can optionally provide ground truth masks, and the rest of the code still looks very similar to the "default" case in the 01 notebook where no mask was provided. Really like the visualizations also of the ground truth and of just running the pre-trained classifier without any training.

Am going to do a review of the code next..

gramhagen:
@PatrickBue did you mean to tag me here?

utils_cv/detection/model.py
```diff
@@ -194,15 +234,21 @@ def __init__(
...
+        self.device = device
+        if self.device is None:
+            self.device = torch_device()
+
         self.model = model
         self.dataset = dataset
         self.im_size = im_size
 
+        if dataset and "labels" in dataset.__dict__:
+            self.labels = dataset.labels
```
PatrickBue: This seems weird. We already store self.dataset which contains self.dataset.labels, and hence should probably not have a second reference via self.labels

utils_cv/detection/model.py
```diff
@@ -236,6 +282,9 @@ def fit(
     ) -> None:
         """ The main training loop. """
 
+        if not self.dataset:
+            return
```
PatrickBue: Should this throw an error rather than returning silently. At the very least this needs a warning message, but throwing an error seems preferred.

utils_cv/detection/model.py
PatrickBue: I agree with JS. This function does not dependent on anything in the function, plus it makes it harder to review since I cannot see what the diff is to before. Should probably remove from here again and place where it used to be. @jiata?

utils_cv/detection/model.py
```diff
@@ -285,63 +334,122 @@ def plot_precision_loss_curves(
...
+        pred = {k: v.detach().cpu().numpy() for k, v in pred.items()}
+        # apply score threshold
+        if threshold:
+            selected = pred['scores'] > threshold
+            pred = {k: v[selected] for k, v in pred.items()}
+        if "masks" in pred and mask_threshold:
+            pred["masks"] = pred["masks"] > mask_threshold
+        return pred
+
+    def _get_det_bboxes_and_mask(
```
PatrickBue: Similar to the function above, i'd suggest to move this function also to where it used to be. This will make it much clearer what the inputs and outputs are (no hidden stuff via self).

utils_cv/detection/model.py
```diff
@@ -285,63 +334,122 @@ def plot_precision_loss_curves(
...
-    ) -> List[DetectionBbox]:
+        **kwargs,
+    ) -> Dict:
         """ Performs inferencing on an image path or image.
 
         Args:
-            im_or_path: the image array which you can get from `Image.open(path)` OR a
-            image path
+            im_or_path: the image array which you can get from
+                `Image.open(path)` OR a image path
```
PatrickBue: [minor] OR should probably be lower case: "or an image path"

utils_cv/detection/model.py
```diff
@@ -285,63 +334,122 @@ def plot_precision_loss_curves(
...
 
-        Raises:
-            TypeError is the im object is a path or str to the image instead of
-            an nd.array
-
         Return a list of DetectionBbox
         """
-        im = (
-            Image.open(im_or_path)
+        im, im_path = (
```
PatrickBue: This seems overly complex. Consider rewriting with "if" clause. Might be 1-2 lines longer, but much easier to understand

utils_cv/detection/model.py
```diff
@@ -285,63 +334,122 @@ def plot_precision_loss_curves(
...
+            Exception: if both `dl` and `self.dataset` are None.
+        """
         if dl is None:
+            if not self.dataset:
+                raise Exception("No dataset provided for evaluation")
             dl = self.dataset.test_dl
         self.results = evaluate(self.model, dl, device=self.device)
         return self.results
 
+    def _transform(
```
PatrickBue: Is this function _transform only ever used once? If so, remove function and copy code where function call used to be.

utils_cv/detection/model.py
```diff
@@ -285,63 +334,122 @@ def plot_precision_loss_curves(
...
         with torch.no_grad():
             pred = model([im])
 
-        labels = self.dataset.labels if self.dataset else self.labels
-        det_bboxes = _get_det_bboxes(pred, labels=labels)
-
-        # limit to threshold if threshold is set
-        return _apply_threshold(det_bboxes, threshold)
+        pred = [
+            self._apply_threshold(p, threshold, **kwargs) for p in pred
```
PatrickBue: Not sure why pred is a list. I.e. we first build a list here, but then in the next line only use pred[0]?

PatrickBue: Given that we only have a single image, why would there be more than 1 item in pred?

utils_cv/detection/model.py
```diff
@@ -353,14 +461,14 @@ def predict_dl(
             threshold: iou threshold for a positive detection. Note: set
             threshold to None to omit a threshold
 
-        Returns a list of DetectionBbox
+        Returns a list of results
         """
         pred_generator = self.predict_batch(dl, threshold=threshold)
-        det_bboxes = [pred for preds in pred_generator for pred in preds]
-        return det_bboxes
+        res = [pred for preds in pred_generator for pred in preds]
```
PatrickBue: Can simplify to return [pred for preds in pred_generator for pred in preds]

utils_cv/detection/model.py
```diff
@@ -373,30 +481,25 @@ def predict_batch(
...
-                )
-
-                det_bboxes = _apply_threshold(det_bboxes, threshold)
-                det_bbox_batch.append(
-                    {"idx": im_idx, "det_bboxes": det_bboxes}
-                )
-            yield det_bbox_batch
+                raw_dets = model(ims)
+
+            raw_dets = [
```
PatrickBue: I am a bit at a loss what is done here. The original version was much easier to read. Can you please undo and use an explicit for xxx in zip(zzz) loop again.

utils_cv/common/plot.py
```diff
@@ -45,25 +45,38 @@ def line_graph(
...
     Args:
-        im_paths (str or List[str]): Image filepaths
+        im_or_im_paths (str or List[str] or numpy.ndarray or List[numpy.ndarray]): Image or image filepaths
         labels (str or List[str]): Image labels. If None, show image file name.
         size (int): MatplotLib plot size.
         rows (int): rows of the images
     """
-    if isinstance(im_paths, (str, Path)):
+    if im_or_im_paths is None or len(im_or_im_paths) == 0:
+        raise Exception(f"Empty {im_or_im_paths}")
```
PatrickBue: Should read: Exception(f"{im_or_im_paths} is either empty or None")

utils_cv/common/plot.py
```diff
@@ -45,25 +45,38 @@ def line_graph(
...
+
+    # im_or_im_paths could be a numpy array of image filepaths
```
PatrickBue: Can this be simplified? I have been looking at this PR for 1.5 hours already, hence getting a bit tired. It's definitely not easy to read. I find it confusing that a single input variable name can either be (i) a single image, (ii) a list of images, (iii) a single image path or (iv) a list of image paths.

PatrickBue:
So, made it through most of the code. I spend 1.5 hours and it's getting difficult to focus. I saw at least one place (maybe more) which might be breaking the hyperdrive notebook.

utils_cv/detection/data.py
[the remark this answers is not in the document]
simonzhaoms: Got it.  I can add it if needed.

utils_cv/detection/model.py
simonzhaoms: Reverted and added mask related code and testing.

utils_cv/detection/dataset.py
```diff
@@ -239,6 +250,24 @@ def _read_annos(self) -> List[str]:
                 self.im_paths.append(im_path)
             else:
                 self.im_paths.append(im_paths[anno_idx])
+
+            if self.mask_dir:
+                # Assume mask image name matches image name but has .png
+                # extension
+                mask_name = os.path.basename(self.im_paths[-1])
+                mask_name = mask_name[:mask_name.rindex('.')] + ".png"
+                mask_name = self.root / self.mask_dir / mask_name
```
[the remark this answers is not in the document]
simonzhaoms: Corrected.

utils_cv/detection/dataset.py
```diff
@@ -386,15 +434,38 @@ def im_gen() -> torch.Tensor:
...
+            if self.mask_paths[idx] is not None:
+                binary_masks = binarise_mask(Image.open(self.mask_paths[idx]))
+            else:
+                # for the tiny bounding box in _read_annos(), make the mask to
+                # be the whole box
+                mask = np.zeros(
+                    Image.open(self.im_paths[idx]).size[::-1],
+                    dtype=np.uint8
+                )
+                mask[1:6, 1:6] = 1
```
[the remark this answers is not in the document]
simonzhaoms: Not sure, since I don't quite understand this part, so I faked the masks in the tiny bounding box.

utils_cv/detection/dataset.py
```diff
@@ -386,15 +434,38 @@ def im_gen() -> torch.Tensor:
...
         Returns a list of annotations and the image path
         """
-        idx = randrange(len(self.anno_paths))
-        return self.anno_bboxes[idx], self.im_paths[idx]
+        idx = random.randrange(len(self.im_paths))
+
+        # get mask if any
+        mask = self._get_binary_masks(idx)
+
+        if mask is not None:
```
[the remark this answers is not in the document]
simonzhaoms: Yes, definitely.  But then `def display_bboxes()` in `utils_cv/detection/plot.py` would be duplicated, so I removed it.

utils_cv/detection/dataset.py
```diff
@@ -386,15 +434,38 @@ def im_gen() -> torch.Tensor:
             print(f"Transformations applied on {self.im_paths[idx]}:")
             [print(transform) for transform in self.transforms.transforms]
 
-    def _get_random_anno(
-        self
-    ) -> Tuple[List[AnnotationBbox], Union[str, Path]]:
+    def _get_binary_masks(self, idx: int) -> Union[np.ndarray, None]:
```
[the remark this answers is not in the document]
simonzhaoms: Done.

utils_cv/detection/model.py
```diff
@@ -88,29 +55,20 @@ def get_pretrained_fasterrcnn(
...
+        model_func: pretrained R-CNN model generating functions, such as
+            fasterrcnn_resnet50_fpn(), get_pretrained_fasterrcnn(), etc.
         min_size: minimum size of the image to be rescaled before feeding it to the backbone
         max_size: maximum size of the image to be rescaled before feeding it to the backbone
         rpn_pre_nms_top_n_train: number of proposals to keep before applying NMS during training
         rpn_pre_nms_top_n_test: number of proposals to keep before applying NMS during testing
         rpn_post_nms_top_n_train: number of proposals to keep after applying NMS during training
         rpn_post_nms_top_n_test: number of proposals to keep after applying NMS during testing
         rpn_nms_thresh: NMS threshold used for postprocessing the RPN proposals
-        box_score_thresh: during inference, only return proposals with a classification score greater than box_score_thresh
```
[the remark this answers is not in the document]
simonzhaoms: No.  they were just refactored into a common function.

utils_cv/detection/model.py
```diff
@@ -88,29 +55,20 @@ def get_pretrained_fasterrcnn(
...
-        box_nms_thresh: NMS threshold for the prediction head. Used during inference
-        box_detections_per_img: maximum number of detections per image, for all classes
 
     Returns
         The model to fine-tine/inference with
```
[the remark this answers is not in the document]
simonzhaoms: Corrected.

utils_cv/detection/model.py
```diff
@@ -133,9 +95,83 @@ def get_pretrained_fasterrcnn(
...
+
+
+def get_pretrained_fasterrcnn(
+    num_classes: int = None,
+    **kwargs,
+) -> nn.Module:
+    """ Gets a pretrained FasterRCNN model
+
+    Args:
+        num_classes: number of output classes of the model (including the background).
```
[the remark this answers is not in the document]
simonzhaoms: Added.

utils_cv/detection/model.py
```diff
@@ -133,9 +95,83 @@ def get_pretrained_fasterrcnn(
...
+    # TODO - reconsider that num_classes includes background. This doesn't feel
+    #     intuitive.
+
+    # load a model pre-trained on COCO
+    model = _get_pretrained_rcnn(
+        fasterrcnn_resnet50_fpn,
+        **kwargs,
+    )
+
+    if num_classes:
```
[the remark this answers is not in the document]
simonzhaoms: Added

utils_cv/detection/model.py
```diff
@@ -123,7 +81,11 @@ def get_pretrained_fasterrcnn(
         box_nms_thresh=box_nms_thresh,
         box_detections_per_img=box_detections_per_img,
     )
+    return model
+
 
+def _tune_box_predictor(model: nn.Module, num_classes: int) -> nn.Module:
```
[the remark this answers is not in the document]
simonzhaoms: Used twice.

utils_cv/detection/model.py
```diff
@@ -133,9 +95,83 @@ def get_pretrained_fasterrcnn(
     return model
 
 
+def _tune_mask_predictor(model: nn.Module, num_classes: int) -> nn.Module:
```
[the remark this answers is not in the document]
simonzhaoms: Removed

utils_cv/detection/model.py
```diff
@@ -133,9 +95,83 @@ def get_pretrained_fasterrcnn(
...
+
+    """
+    # load a model pre-trained on COCO
+    model = _get_pretrained_rcnn(
+        maskrcnn_resnet50_fpn,
+        **kwargs,
+    )
+
+    if num_classes:
+        model = _tune_box_predictor(model, num_classes)
```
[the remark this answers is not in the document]
simonzhaoms: not understand?

utils_cv/detection/model.py
simonzhaoms: We allow `self.dataset is None`, so `add_labels()` is added to provide labels.

utils_cv/detection/model.py
simonzhaoms: Done.

utils_cv/detection/model.py
simonzhaoms: corrected.

utils_cv/detection/model.py
simonzhaoms: Rewritten.

utils_cv/detection/model.py
simonzhaoms: Removed

utils_cv/detection/model.py
simonzhaoms: The pre-trained PyTorch mask rcnn model needs inputs as list even if there is only one image.  I have made it simpler.

utils_cv/detection/model.py
simonzhaoms: Done

utils_cv/detection/model.py
simonzhaoms: Yeah, a little bit.  Rewritten.

utils_cv/common/plot.py
simonzhaoms: Done.

utils_cv/common/plot.py
simonzhaoms: Yeah, reverted.

utils_cv/detection/dataset.py
```diff
@@ -239,6 +250,24 @@ def _read_annos(self) -> List[str]:
...
+                # Assume mask image name matches image name but has .png
+                # extension
+                mask_name = os.path.basename(self.im_paths[-1])
+                mask_name = mask_name[:mask_name.rindex('.')] + ".png"
+                mask_name = self.root / self.mask_dir / mask_name
+                # For mask prediction, if no mask provided and negatives not
+                # allowed (), ignore the image
+                if not mask_name.exists():
+                    if os.path.exists(anno_path):
+                        del self.im_paths[-1]
```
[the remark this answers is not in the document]
simonzhaoms: Yeah.  Agree.  Done.

utils_cv/detection/model.py
simonzhaoms: Reverted with mask related code.

utils_cv/detection/dataset.py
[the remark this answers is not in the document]
PatrickBue: Let's assume for now that all images have masks. If not, then an error should be thrown during dataset creation. WIth that assumption, the else can get removed.

utils_cv/detection/model.py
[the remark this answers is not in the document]
PatrickBue: Sounds good!

utils_cv/detection/model.py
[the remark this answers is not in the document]
PatrickBue: I think I was wrong. I assumed that "_tune_box_predictor()" does not need to get called since "_tune_mask_predictor()" was called. But guess both have to be called indeed. If so then ignore this comment.

utils_cv/detection/model.py
PatrickBue: I see. In your notebook, it uses add_labels because not dataset was specified:
   detector = DetectionLearner(model=get_pretrained_maskrcnn(), device=torch.device(DEVICE))
   detector.add_labels(coco_labels()[1:])

If add_labels HAS to be called (which I believe is the case) if no dataset was specified, then I would add a "labels" variable to the class constructor. And in the constructor it can throw an error if no dataset was provided and no "labels" variable.

utils_cv/detection/model.py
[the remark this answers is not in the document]
PatrickBue: Ah, I found the parameters. Looks like these 3 lines of doc text describing box_score_thresh, box_nms_thresh and box_detections_per_img got erroneously removed? The variables are still there, just no this text.

utils_cv/detection/dataset.py
```diff
@@ -386,15 +428,35 @@ def im_gen() -> torch.Tensor:
...
         """ Get random annotation and corresponding image
 
         Returns a list of annotations and the image path
         """
-        idx = randrange(len(self.anno_paths))
-        return self.anno_bboxes[idx], self.im_paths[idx]
+        idx = random.randrange(len(self.im_paths))
+
+        # get mask if any
+        mask = self._get_binary_mask(idx)
```
PatrickBue: Could just add this on the same line below, ie.
return self.anno_bboxes[idx], self.im_paths[idx], self._get_binary_mask(idx)

utils_cv/detection/model.py
```diff
@@ -23,54 +35,59 @@
...
-            box,
-            score=score,
-            label_idx=label,
-            label_name=label_name,
-            im_path=im_path,
-        )
-        det_bboxes.append(det_bbox)
 
-    return det_bboxes
+    label_names = np.array(labels)[pred['labels'] - 1]
```
PatrickBue: The function looks nice since statements like "pred_labels = pred[0]["labels"].detach().cpu().numpy().tolist()" got removed. But now it's assumed that the _apply_threshold() function HAS to be called, since in there the detach() etc is called. This is not nice. My suggestion: change back to the original code and remove the from_arrays() function since I guess it's not used anymore.

utils_cv/detection/model.py
```diff
@@ -23,54 +35,59 @@
...
+                im_path=im_path,
+            )
+    }
+
+    if "masks" in pred:
+        res["masks"] = pred["masks"].squeeze(1)
+    return res
 
 
 def _apply_threshold(
```
PatrickBue: Similar to the comment above, apply_threshold() in it's current form does more than just applying threshold: it also calls detach() etc. Now this function always has to be called. Please revert back to before where it just returns a subset of detections.

utils_cv/detection/model.py
```diff
@@ -23,54 +35,59 @@
...
-    det_bboxes: List[DetectionBbox], threshold: Optional[float] = 0.5
-) -> List[DetectionBbox]:
-    """ Filters the list of DetectionBboxes by score threshold. """
-    return (
-        [det_bbox for det_bbox in det_bboxes if det_bbox.score > threshold]
-        if threshold is not None
-        else det_bboxes
-    )
+    pred: Dict,
```
PatrickBue: Similar to the comment above, apply_threshold() in it's current form does more than just applying threshold: it also calls detach() etc. Now this function always has to be called. Please revert back to before where it just returns a subset.

PatrickBue: I believe we probably want to get rid of "mask_threshold". "threshold" makes sense since it removes a detection is the confidence in the FULL detection is lower than a threshold (not just in the mask)

utils_cv/detection/model.py
PatrickBue: I see. Maybe just do pred = model([im])[0] then. In one line we create a list of im, and then take the 0th element.

utils_cv/detection/model.py
```diff
@@ -309,39 +421,41 @@ def predict(
         self,
         im_or_path: Union[np.ndarray, Union[str, Path]],
         threshold: Optional[int] = 0.5,
-    ) -> List[DetectionBbox]:
+        **kwargs,
```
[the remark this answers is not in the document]
PatrickBue: See also my comment much earlier that likely we don't want "mask_threshold" at all.

utils_cv/detection/plot.py
```diff
@@ -124,6 +150,44 @@ def display_bboxes(
         plt.show()
 
 
+def display_bbox_mask(
```
PatrickBue: Should be called "display_bboxes_mask", similar to what it was called before since it takes multiple bboxes as input

utils_cv/detection/plot.py
```diff
@@ -86,32 +89,55 @@ def plot_boxes(
...
+    for cmask in colored_masks:
+        tmask = Image.fromarray(
+            transparentise_mask(cmask, plot_settings.mask_alpha)
+        )
+        im = Image.alpha_composite(im, tmask)
+
+    return im
+
+
+def display_image(
```
PatrickBue: If only used once, then please delete this function and move code.

utils_cv/detection/bbox.py
```diff
@@ -25,14 +30,66 @@ def __init__(self, left: int, top: int, right: int, bottom: int):
...
-    def from_array(cls, arr: List[int]) -> "Bbox":
+    def from_array(cls, arr: List[int]) -> "_Bbox":
         """ Create a Bbox object from an array [left, top, right, bottom] """
         return _Bbox(arr[0], arr[1], arr[2], arr[3])
 
     @classmethod
-    def from_array_xywh(cls, arr: List[int]) -> "Bbox":
-        """ create a Bbox object from an array [left, top, width, height] """
-        return _Bbox(arr[0], arr[1], arr[0] + arr[2], arr[1] + arr[3])
+    def from_array_xywh(cls, arr: List[int]) -> "_Bbox":
```
PatrickBue: There are a lot new helper functions in here. If they are indeed used, then that's ok. If they are not used, then please delete.

utils_cv/detection/bbox.py
```diff
@@ -25,14 +30,66 @@ def __init__(self, left: int, top: int, right: int, bottom: int):
...
+        width: int,
+    ) -> Union[torch.Tensor, np.ndarray, List[List[int]]]:
+        """ Flip rectangles horizontally.
+
+        Args:
+            rects: list of rectangles in the form of [left, top, right, bottom]
+            width: width of the image which the rectangles are from
+        """
+        islist = False
+        if isinstance(rects, list):
```
PatrickBue: Does this function need to support list and np.array, ie is this functionality used somewhere? If not, then remove code that's not needed to simplify this function.

utils_cv/detection/bbox.py
```diff
@@ -112,6 +169,20 @@ def is_valid(self) -> bool:
             return False
         return True
 
+    def hflip(self, width) -> "_Bbox":
```
PatrickBue: Are those two functions used anywhere? If not then can remove.

utils_cv/detection/bbox.py
```diff
@@ -145,6 +216,64 @@ def from_array(cls, arr: List[int], **kwargs) -> "AnnotationBbox":
         bbox.set_meta(**kwargs)
         return bbox
 
+    @classmethod
+    def from_array_xywh(cls, arr: List[int], **kwargs) -> "AnnotationBbox":
```
PatrickBue: I don't feel good about the **kwargs, and generally about so many new helper functions. Can you please remove all functions which are not used and I will then look at them again.

scenarios/detection/12_hard_negative_sampling.ipynb
```diff
@@ -548,7 +548,7 @@
    ],
    "source": [
     "# Get validation accuracy on test set at IOU=0.5:0.95\n",
-    "acc = float(detector.ap[-1])\n",
```
PatrickBue: This edit also needs to be done in the 11_ notebook since otherwise the 11_ notebook will start failing (at least) on this line:
     acc = float(detector.ap[-1])

This problem actually comes from the fact that our AML notebook testing devops pipeline is currently broken.

PatrickBue:
@simonzhaoms Thanks for making some of the edits already! I now went through all the scripts (except for the tests scripts) and left more comments. 

The code looks pretty good already, and my comments are mostly small items. However there are many of them which is due to this PR being massive.

I'd recommend to first address the comments, then to ask for me to look at the code again, and only afterwards to fix the tests again (that should make you life easier).

utils_cv/detection/model.py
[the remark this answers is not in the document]
simonzhaoms: :smile:

utils_cv/detection/model.py
simonzhaoms: Make sense.  @jiata What do you think?

utils_cv/detection/model.py
simonzhaoms: Changed.

utils_cv/detection/dataset.py
simonzhaoms: Done

utils_cv/detection/plot.py
simonzhaoms: Renamed

utils_cv/detection/plot.py
simonzhaoms: Removed

utils_cv/detection/bbox.py
simonzhaoms: Removed

utils_cv/detection/bbox.py
simonzhaoms: All removed.  Those functions are legacy code which I forgot to remove, and I also added tests for them :man_facepalming:

utils_cv/detection/bbox.py
simonzhaoms: Deleted.

utils_cv/detection/bbox.py
simonzhaoms: Removed this function

scenarios/detection/12_hard_negative_sampling.ipynb
simonzhaoms: Thanks.  Done.  Things getting tedious when code modification involves other collaborators' work :man_facepalming:

utils_cv/detection/model.py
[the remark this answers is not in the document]
simonzhaoms: Changed.

utils_cv/detection/model.py
simonzhaoms: Reverted and removed `from_arrays()`

utils_cv/detection/model.py
simonzhaoms: Do you mean to hard-code `mask_threshold` to `0.5` where it appears?

utils_cv/detection/dataset.py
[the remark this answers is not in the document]
simonzhaoms: Removed and all tests are passed currently.

utils_cv/detection/model.py
simonzhaoms: Changed.

utils_cv/detection/model.py
simonzhaoms: I hard-coded `mask_threshold` to `0.5`

PatrickBue: Thanks. This seems like a good approach for now to clean up this PR.

utils_cv/detection/bbox.py
PatrickBue: Thanks for deleting most of the new helper functions. This looks much nicer now and is also easier to maintain going forward

scenarios/detection/12_hard_negative_sampling.ipynb
PatrickBue: @simonzhaoms Yeah, fully understand, I would have the same problem. The main cause of this problem is that the PR is so big. Good advice / painful learning to always try to add small bits of code which solve sub-problems rather than one big PR which solves the full problem.

tests/integration/detection/test_integration_detection.py
```diff
@@ -23,9 +23,40 @@ def test_01_notebook_run(detection_notebooks):
...
+def test_02_notebook_run(detection_notebooks):
+    epochs = 5
+    notebook_path = detection_notebooks["02"]
+    pm.execute_notebook(
+        notebook_path,
+        OUTPUT_NOTEBOOK,
+        parameters=dict(
+            PM_VERSION=pm.__version__,
+            EPOCHS=epochs,
+            DEVICE='cuda'
```
PatrickBue: [minor] I believe this line device='cuda' is not needed anymore.

tests/unit/detection/test_detection_notebooks.py
```diff
@@ -47,9 +47,37 @@ def test_01_notebook_run(detection_notebooks, tiny_od_data_path):
...
+@pytest.mark.notebooks
+def test_02_notebook_run(detection_notebooks, tiny_od_mask_data_path):
+    notebook_path = detection_notebooks["02"]
+    pm.execute_notebook(
+        notebook_path,
+        OUTPUT_NOTEBOOK,
+        parameters=dict(
+            PM_VERSION=pm.__version__,
+            DATA_PATH=tiny_od_mask_data_path,
+            DEVICE='cuda',
```
PatrickBue: [minor] no need to specify DEVICE here

utils_cv/detection/dataset.py
```diff
@@ -386,15 +428,30 @@ def im_gen() -> torch.Tensor:
...
-        self
-    ) -> Tuple[List[AnnotationBbox], Union[str, Path]]:
+    def _get_binary_mask(self, idx: int) -> Union[np.ndarray, None]:
+        """ Return binary masks for objects in the mask image. """
+        binary_masks = None
+        if self.mask_paths:
+            if self.mask_paths[idx] is not None:
+                binary_masks = binarise_mask(Image.open(self.mask_paths[idx]))
+            else:
+                # for the tiny bounding box in _read_annos(), make the mask to
```
PatrickBue: [minor] this comment seems out-dated

PR #415, opened 2019-11-08
Count based OD evaluation

PatrickBue:
### Description
Adding counting based evaluation measure.

### Checklist:
- [x ] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [x ] I have added tests covering my contributions.
- [x ] I have updated the documentation accordingly.


utils_cv/detection/bbox.py
```diff
@@ -194,3 +194,27 @@ def from_array(
 
     def __repr__(self):
         return f"{super().__repr__()} | score: {self.score}"
+
+
+def boxes_iou(box1: DetectionBbox, box2: DetectionBbox):
```
PatrickBue: I am calling all variables "boxes" rather than "bboxes". Is that OK? Should we use "boxes" going forward everywhere instead of the longer "bboxes"?

PatrickBue: Or if in code we use "bboxes" (nearly) everywhere should I rename this also to "bboxes"?

PatrickBue:
@jiata Note that in the notebook https://github.com/microsoft/ComputerVision/blob/pabuehle_counts/scenarios/detection/01_training_introduction.ipynb detector.predict_dl(data.test_dl) is called twice, once with detector.predict_dl(data.test_dl, threshold=0) and once with detector.predict_dl(data.test_dl, threshold=0.5). This is not optimal. Let me know if you have any suggestion how to avoid that.

utils_cv/detection/bbox.py
jiata: i think bboxes for consistency is preferable. its also more obviously "bounding box" in my opinion.

jiata:
> @jiata Note that in the notebook https://github.com/microsoft/ComputerVision/blob/pabuehle_counts/scenarios/detection/01_training_introduction.ipynb detector.predict_dl(data.test_dl) is called twice, once with detector.predict_dl(data.test_dl, threshold=0) and once with detector.predict_dl(data.test_dl, threshold=0.5). This is not optimal. Let me know if you have any suggestion how to avoid that.

I *think* there might be a way to get there data we need from the evaluator object. `bbox = e.eval_imgs["bbox"]` 

This is how i implemented the PR curve. Instead of calling predict() again, i was able to just use the eval object.

PR #454, opened 2019-12-09
Add keypoint detection with tuned model

simonzhaoms:
### Description

Add keypoint model tuning for custom datasets

### Checklist:


PatrickBue:
@simonzhaoms I just had a look at the notebook. Looks very good already.

utils_cv/detection/dataset.py
```diff
@@ -107,10 +150,44 @@ def parse_pascal_voc_anno(
...
         label = obj.find("name").text
+        # Get keypoints if any
+        if keypoint_meta is not None:
+            kps = []
+            kps_category = keypoint_meta["category"]
+            kps_labels = keypoint_meta["labels"]
+
+            # It seems Keypoint R-CNN model doesn't accept varied number of
+            # keypoints, so we only allow one category currently.
+            if label != kps_category:
```
PatrickBue: Not sure we want to implement this. Seems odd to have a dataset and then only read some of the images. The likely better alternative is to create a subset of the dataset with just e.g. milk bottle images. See also my comment in the notebook. Furthermore, having multiple objects I believe is ok IF they have the same number of keypoints.

utils_cv/detection/dataset.py
```diff
@@ -20,12 +20,48 @@
...
-from .references.transforms import RandomHorizontalFlip, Compose, ToTensor
+from .references.transforms import Compose, ToTensor
 from ..common.gpu import db_num_workers
 
 Trans = Callable[[object, dict], Tuple[object, dict]]
 
 
+def _flip_keypoints(kps, width, hflip_inds):
+    """ Variation of `references.transforms._flip_coco_person_keypoints` with additional
+    hflip_inds. """
```
PatrickBue: See my comment in the notebook. Do we really need hflip_inds? And if not, does that mean that this function will not be needed anymore?

utils_cv/detection/dataset.py
```diff
@@ -20,12 +20,48 @@
...
 from .mask import binarise_mask
 from .references.utils import collate_fn
-from .references.transforms import RandomHorizontalFlip, Compose, ToTensor
+from .references.transforms import Compose, ToTensor
 from ..common.gpu import db_num_workers
 
 Trans = Callable[[object, dict], Tuple[object, dict]]
 
 
+def _flip_keypoints(kps, width, hflip_inds):
```
PatrickBue: [minor] kps should probably be called keypoints for consistency

utils_cv/detection/dataset.py
```diff
@@ -20,12 +20,48 @@
...
+    """ Variation of `references.transforms._flip_coco_person_keypoints` with additional
+    hflip_inds. """
+    flipped_data = kps[:, hflip_inds]
```
PatrickBue: [minor] flipped_data should probably be called flipped_keypoints for consistency

utils_cv/detection/dataset.py
```diff
@@ -20,12 +20,48 @@
...
+
+
+class RandomHorizontalFlip(object):
+    """ Variation of `references.transforms.RandomHorizontalFlip` to make sure flipping
+    works on custom keypoints. """
+
+    def __init__(self, prob):
+        self.prob = prob
+
+    def __call__(self, image, target):
```
PatrickBue: [minor] image should probably be called "im" for consistency

utils_cv/detection/dataset.py
```diff
@@ -223,12 +304,22 @@ def _read_annos(self) -> None:
...
+                anno_bboxes, im_path, keypoints = parse_pascal_voc_anno(
+                    anno_path, keypoint_meta=self.keypoint_meta
+                )
+                # When meta provided, we assume this is keypoint
+                # detection, and skip the image if no keypoints found.
+                if self.keypoint_meta is not None:
+                    if len(keypoints) != 0:
+                        self.keypoints.append(keypoints)
+                    else:
+                        continue
```
PatrickBue: So the list self.keypoints could be shorted than self.im_paths, which would be a bug? Should we throw an error here - if the user specifies keypoint_meta, then all images need to have the keypoint annotation for training?

tests/conftest.py
```diff
@@ -560,6 +573,63 @@ def od_detection_mask_dataset(tiny_od_mask_data_path):
     )
 
 
+@pytest.fixture(scope="session")
+def carton_keypoint_meta():
+    return {
+        "category": "carton",
+        "skeleton": [
```
PatrickBue: [minor] See my comment from the notebook. When defining a COCOkeypoint_meta class (or whatever name) then can just use here rather than defining below.

utils_cv/detection/dataset.py
simonzhaoms: Generated a dataset with milk bottle.  Yes, it is OK to have multiple objects of the same category in a image.

utils_cv/detection/dataset.py
simonzhaoms: In `utils_cv.detection.dataset.get_transform()`, `RandomHorizontalFlip` is used to flip the image, so `hflip_inds` is necessary.

utils_cv/detection/dataset.py
simonzhaoms: Renamed

utils_cv/detection/dataset.py
simonzhaoms: Done

utils_cv/detection/dataset.py
simonzhaoms: Done

utils_cv/detection/dataset.py
simonzhaoms: Changed.  And the exception will be thrown when parsing the annotation xml file in `utils_cv.detection.dataset.parse_pascal_voc_anno()`

tests/conftest.py
simonzhaoms: Added `carton_keypoint_meta` in `keypoint.py`

utils_cv/detection/dataset.py
simonzhaoms: In training, the data would be augmented
https://github.com/microsoft/computervision-recipes/blob/8602f5ce79e5466ab15c1a69f17f82c5c87bd0fb/utils_cv/detection/dataset.py#L511-L512
by some kind of transformations:
https://github.com/microsoft/computervision-recipes/blob/8602f5ce79e5466ab15c1a69f17f82c5c87bd0fb/utils_cv/detection/dataset.py#L49-L79
`RandomHorizontalFlip` is one of the transformations used in training:
https://github.com/microsoft/computervision-recipes/blob/8602f5ce79e5466ab15c1a69f17f82c5c87bd0fb/utils_cv/detection/references/transforms.py#L26-L43
For generating flipped keypoint image data:
https://github.com/microsoft/computervision-recipes/blob/8602f5ce79e5466ab15c1a69f17f82c5c87bd0fb/utils_cv/detection/references/transforms.py#L6-L13
'left' would become 'right' after flipped horizontally.  Therefore, to get `DetectionDataset`work properly for keypoint detection in its `__getitem__()` method
https://github.com/microsoft/computervision-recipes/blob/8602f5ce79e5466ab15c1a69f17f82c5c87bd0fb/utils_cv/detection/dataset.py#L468
we need `hflip_inds` to indicate how to transform the keypoint data correctly.

Anyway, we can also leave it to users to define how to transform the data by themselves.

utils_cv/detection/dataset.py
```diff
@@ -154,17 +154,14 @@ def parse_pascal_voc_anno(
     objs = root.findall("object")
     for obj in objs:
         label = obj.find("name").text
-        # Get keypoints if any
+        # Get keypoints if any.
+        # For keypoint detection, currently only one category (except
```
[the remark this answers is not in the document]
simonzhaoms: Yes, only one category is allowed.  For example milk-bottle detector can only be used to detect milk bottles and their predefined keypoints in the images.  If water bottle needs to be detected, another water-bottle detector should be trained first.

utils_cv/detection/keypoint.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+COCO_keypoint_meta = {
```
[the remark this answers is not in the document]
simonzhaoms: Moved into notebook

utils_cv/detection/dataset.py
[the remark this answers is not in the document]
PatrickBue: If the categories have the same keypoints (e.g. "top" and "bottom") would then be multiple categories allowed?

utils_cv/detection/dataset.py
PatrickBue: I understand, thanks for the explanation. Since hflip_inds is a bit tricky to get right, would it be possible to allow it to be None (by default) and in that case the RandomHorizontalFlip() would throw an error with message something like "To use random horizontal flipping, hflip_inds needs to be specified"

utils_cv/detection/keypoint.py
```diff
@@ -45,58 +45,3 @@
     # left becomes right when flipped horizontally
     "hflip_inds": [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15],
```
PatrickBue: [minor] Would you mind adding another 1-2 sentences here with explanation what hflip_inds does, how it works, what these numbers mean (I understand now, but somebody who's new to this wouldn't)

simonzhaoms: Done

utils_cv/detection/dataset.py
simonzhaoms: Done

PatrickBue:
@simonzhaoms Thanks for making the last edits. This PR looks great now - feel free to merge

utils_cv/detection/dataset.py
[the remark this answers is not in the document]
simonzhaoms: Yes.  I will add that in another PR.

PR #458, opened 2019-12-19
Plot keypoints as round dots to make them noticeable

simonzhaoms:
### Description

Plot keypoints as round dots to make them noticeable.

### Checklist:

### DIFF
--- a/utils_cv/detection/plot.py
+++ b/utils_cv/detection/plot.py
@@ -171,6 +171,16 @@ def plot_keypoints(
                 width=plot_settings.keypoint_th,
             )
 
+        # draw keypoints
+        visible_point_xys = keypoints[keypoints[..., 2] != 0][..., :2]
+        offset = 2 * plot_settings.keypoint_th
+        point_bboxes = np.hstack([
+            visible_point_xys - offset,  # left top
+            visible_point_xys + offset,  # right bottom
+        ])
+        for bbox in point_bboxes.tolist():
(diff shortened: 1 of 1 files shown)


utils_cv/detection/plot.py
```diff
@@ -171,6 +171,16 @@ def plot_keypoints(
                 width=plot_settings.keypoint_th,
             )
 
+        # draw keypoints
+        visible_point_xys = keypoints[keypoints[..., 2] != 0][..., :2]
+        offset = 2 * plot_settings.keypoint_th
+        point_bboxes = np.hstack([
```
PatrickBue: "bbox" seems like the wrong variable name here, since we use it to refer to an instance of the BoundingBox class. Maybe "rect" instead for rectangle?

utils_cv/detection/plot.py
```diff
@@ -171,6 +171,16 @@ def plot_keypoints(
...
+            visible_point_xys - offset,  # left top
+            visible_point_xys + offset,  # right bottom
+        ])
+        for bbox in point_bboxes.tolist():
```
PatrickBue: same here. E.g. "for rect in rects"

utils_cv/detection/plot.py
simonzhaoms: Renamed

utils_cv/detection/plot.py
simonzhaoms: Done

PR #473, opened 2020-01-17
update video annotation code

angusrtaylor:
### Description

Added functionality to the video annotation tools to
- Generate randomly sampled negative examples
- Split dataset into training and test sets, with option to create sets for hard negative mining


environment.yml
```diff
@@ -17,8 +17,8 @@ channels:
 - fastai
 dependencies:
 - python==3.6.8
-- pytorch>=1.2.0
-- torchvision>=0.3.0
+- pytorch==1.2.0
+- torchvision==0.4.0
```
PatrickBue: Can we undo this change, i.e. not edit environment.yml (since that might break our devops tests) @angusrtaylor

PR #477, opened 2020-01-27
Video Dataset / Model refactor + framework for action tests

jiata:
### Description
basic framework for testing for action recognition 

### Checklist:


jiata:
still a lot of work needs to be done for testing.. ended up getting stuck with some environment issues (so you'll notice that the environment.yml is also updated). but hopefully this starts to get at the basic skeleton for the tests.

PatrickBue:
Hi JS! I should be able to start working on this next week. Shall I use your branch, or should I branch off from your branch - any preference?

jiata:
> Hi @jiata ! I should be able to start working on this next week. Shall I use your branch, or should I branch off from your branch - any preference?

Feel free to use my branch directly if that's easier!

PatrickBue:
> Hi @jiata ! I should be able to start working on this next week. Shall I use your branch, or should I branch off from your branch - any preference?
> 
> Feel free to use my branch directly if that's easier!

Thanks, I will do that then.

environment.yml
```diff
@@ -40,4 +40,7 @@ dependencies:
   - nvidia-ml-py3
   - nteract-scrapbook
   - azureml-sdk[notebooks,contrib]>=1.0.30
+  - decord==0.3.5
+  - einops==0.1.0
+  - numpy==1.17.4 # https://github.com/pytorch/vision/issues/1700#issuecomment-569327090
```
PatrickBue: - Would be better to install these new libraries in the conda, rather than pip, section to avoid dependency problems. I had a look though and seems that both decord and einops are pip only.
- Young made an update to the cocoapi which hopefully means we don't need to revert numpy to a specific version. I am afraid this would give us headaches in the future.

youngpark: +1

tests/conftest.py
```diff
@@ -79,39 +91,33 @@ def classification_notebooks():
...
-            folder_notebooks, "01_training_introduction.ipynb"
-        ),
-        "02_multilabel_classification": os.path.join(
+        "00": os.path.join(folder_notebooks, "00_webcam.ipynb"),
+        "01": os.path.join(folder_notebooks, "01_training_introduction.ipynb"),
+        "02": os.path.join(
             folder_notebooks, "02_multilabel_classification.ipynb"
         ),
-        "03_training_accuracy_vs_speed": os.path.join(
+        "03": os.path.join(
```
PatrickBue: Thanks for making these edits, also noticed that before!

scenarios/action_recognition/02_training_hmdb.ipynb
```diff
@@ -0,0 +1,965 @@
+{
```
PatrickBue: <p>[minor] seems a bit odd if it's a batch to include the number of samples. Don't think fast.ai does it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/computervision-recipes/pull/477/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

PatrickBue: <p>Really like how this and the VideoDataset follows the detection naming and syntax.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/computervision-recipes/pull/477/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+            [VideoRecord(x.strip().split(" ")) for x in open(test_split_file)]
+        )
+
+        # create indices
+        indices = torch.arange(0, len(self.video_records))
+        train_range = indices[:train_len]
+        test_range = indices[train_len:]
+
+        # create train subset
+        train = copy.deepcopy(Subset(self, train_range))
```
PatrickBue: [question] Remind me, we do these deepcopies to create a structure similar to fast.ai?

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+
+    def init_data_loaders(self) -> None:
+        """ Create training and validation data loaders. """
+        devices = num_devices()
+
+        self.train_dl = DataLoader(
+            self.train_ds,
+            batch_size=self.batch_size * devices,
+            shuffle=True,
+            num_workers=0,
```
PatrickBue: Instead of num_workers=0, I guess we want to use: num_workers=db_num_workers(),

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+            batch_size=self.batch_size * devices,
+            shuffle=False,
+            num_workers=0,
+            pin_memory=True,
+        )
+
+    def __len__(self) -> int:
+        return len(self.video_records)
+
+    def _sample_indices(self, record: VideoRecord) -> List[int]:
```
PatrickBue: Some more info what this function does would be great

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+                    cur_frame = video_reader.next().asnumpy()
+                    if len(cur_frame.shape) != 3:
+                        # maybe end of the video
+                        break
+                    clip.append(cur_frame)
+        except StopIteration:
+            pass
+
+        # if clip needs more frames, simply duplicate the last frame in the clip.
+        while len(clip) < self.sample_length:
```
PatrickBue: How often is this the case? If not common then totally fine. Ff more common, is that the right approach?

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+                            # maybe end of the video
+                            break
+                        clip.append(cur_frame)
+            else:
+                for i in range(self.sample_length - 1):
+                    if self.sample_step > 1:
+                        video_reader.skip_frames(self.sample_step - 1)
+                    cur_frame = video_reader.next().asnumpy()
+                    if len(cur_frame.shape) != 3:
+                        # maybe end of the video
```
PatrickBue: This is probably not code you wrote but do you know it this ever happens, and is it really "maybe end of video"

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+                    else:
+                        if step > 1:
+                            video_reader.skip_frames(step - 1)
+                        cur_frame = video_reader.next().asnumpy()
+                        if len(cur_frame.shape) != 3:
+                            # maybe end of the video
+                            break
+                        clip.append(cur_frame)
+            else:
+                for i in range(self.sample_length - 1):
```
PatrickBue: Is this here just a special case of the self.temporal_jitter, just with no jitter, and hence could be handled by the code above? The two code blocks seem to have a good number of code lines in common.

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+                    if self.sample_step > 1:
+                        video_reader.skip_frames(self.sample_step - 1)
+                    cur_frame = video_reader.next().asnumpy()
+                    if len(cur_frame.shape) != 3:
+                        # maybe end of the video
+                        break
+                    clip.append(cur_frame)
+        except StopIteration:
```
PatrickBue: [minor] add comment here when this exception is thrown (maybe once the end of the video is reached?)

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+                ax = [ax]
+            for j, a in enumerate(ax):
+                a.axis("off")
+                a.imshow(
+                    np.moveaxis(denormalize(clip[j], mean, std).numpy(), 0, -1)
+                )
+            pass
+
+    def show_batch(
+        self, train_or_test: str = "train", num_samples: int = 1
```
PatrickBue: [minor] fyi, fast.ai uses 'rows' instead of 'num_samples'

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+        """Plot first few samples in the datasets"""
+        if train_or_test == "train":
+            batch = [self.train_ds.dataset[i][0] for i in range(num_samples)]
+        elif train_or_test == "valid":
+            batch = [self.test_ds.dataset[i][0] for i in range(num_samples)]
+        else:
+            raise ValueError("Unknown data type {}".format(which_data))
+
+        self._show_batch(
+            batch, self.sample_length, mean=DEFAULT_MEAN, std=DEFAULT_STD,
```
PatrickBue: [minor] dont think this is necessary: mean=DEFAULT_MEAN, std=DEFAULT_STD

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+        batch_size = len(batch)
+        plt.tight_layout()
+        fig, axs = plt.subplots(
+            batch_size,
+            sample_length,
+            figsize=(4 * sample_length, 3 * batch_size),
+        )
+
+        for i, ax in enumerate(axs):
+            if batch_size == 1:
```
PatrickBue: Is this line necessary, and if so why?

utils_cv/action_recognition/model.py
```diff
@@ -233,37 +175,36 @@ def fit(self, train_cfgs):
             )
 
         # Learning rate scheduler
-        if train_cfgs.get('use_one_cycle_policy', False):
+        if train_cfgs.get("use_one_cycle_policy", False):
```
PatrickBue: Fyi, the parameters I found to work well for HMDB-51 (and now already used in the master 01 notebook) were for simple StepLR because our older pytorch version does not support OneCycleLR. Should we update pytorch we could re-visit and see if OneCycleLR performs better (but would have to tune at least the learning rate then again).

utils_cv/action_recognition/model.py
```diff
@@ -233,37 +175,36 @@ def fit(self, train_cfgs):
...
-            model = self.model
+        model = (
+            nn.DataParallel(self.model) if count_devices > 1 else self.model
+        )
 
         criterion = nn.CrossEntropyLoss().to(device)
 
         for e in range(1, train_cfgs.epochs + 1):
             print("Epoch {} ==========".format(e))
             if scheduler is not None:
```
PatrickBue: I believe there is no need to check if scheduler is None here

utils_cv/action_recognition/model.py
```diff
@@ -296,23 +239,26 @@ def train_an_epoch(
...
+            dict {
+                'train/time': batch_time.avg,
+                'train/loss': losses.avg,
+                'train/top1': top1.avg,
+                'train/top5': top5.avg,
+                'valid/time': ...
+            }
         """
         assert "train" in data_loaders
         if mixed_prec and not AMP_AVAILABLE:
```
PatrickBue: Similar assert was already earlier. Could remove or delete one.

utils_cv/action_recognition/model.py
```diff
@@ -296,23 +239,26 @@ def train_an_epoch(
...
+        Return:
+            dict {
+                'train/time': batch_time.avg,
+                'train/loss': losses.avg,
+                'train/top1': top1.avg,
+                'train/top5': top5.avg,
+                'valid/time': ...
+            }
         """
         assert "train" in data_loaders
```
PatrickBue: [minor] could maybe remove this, or also add that "valid" is in dataloaders.

utils_cv/action_recognition/model.py
```diff
@@ -357,7 +303,9 @@ def train_an_epoch(
                         loss = loss / grad_steps
 
                         if mixed_prec and AMP_AVAILABLE:
```
PatrickBue: Same as above, should only probably check for mixed_prec and not silently ignore if AMP isn't available.

scenarios/action_recognition/02_training_hmdb.ipynb
jiata: its the same things as the `rows` parameter in a way - maybe i should rename to that?

utils_cv/action_recognition/dataset.py
jiata: hmm dont remember how it related to fastai. but for object detection, we used the same pattern so that the dataset object could hold separate copies for the train and validation sets, each with their own parameters.

utils_cv/action_recognition/dataset.py
jiata: Torch 1.2 has a bug when num-workers > 0 (0 means run a main-processor worker) - added this as a note

utils_cv/action_recognition/dataset.py
jiata: based on some quick testing, this happens extremely infrequently using the hmdb51 dataset

utils_cv/action_recognition/dataset.py
jiata: based on some quick tests, it never gets to this condition. so i'm going to remove it.

utils_cv/action_recognition/dataset.py
jiata: refactored

utils_cv/action_recognition/dataset.py
jiata: unfortunately this is necessary bc we're iterating through the axs which when rows=1, is the size of the number of columns. But when rows>1,  is the size of the rows.

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+    Args:
+        train: whether or not this is for training
+        tfms_config: Config object with tranforms-related configs
+
+    Returns:
+        A list of transforms to apply
+    """
+    if tfms_config is None:
+        tfms_config = (
+            get_default_tfms_config(train=True)
```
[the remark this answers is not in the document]
jiata: feels more clear with `train=True` here

utils_cv/common/misc.py
```diff
@@ -81,3 +81,34 @@ def get_font(size: int = 12) -> ImageFont:
             font = None
 
     return font
+
+
+class Config(object):
```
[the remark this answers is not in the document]
jiata: i so didnt write the function, but i decided to keep it bc this pattern is used throughout the 'model.py' file:
```
config_object.get("field", <default_value>)
```

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+            random_crop=random_crop,
+            random_crop_scales=random_crop_scales,
+        )
+    )
+
+
+class VideoDataset:
+    """ A video recognition dataset. """
+
+    def __init__(
```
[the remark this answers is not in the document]
jiata: hrmm. lets think on this (maybe its a utility that we can implement generically across the repos)

utils_cv/action_recognition/model.py
```diff
@@ -212,17 +154,17 @@ def fit(self, train_cfgs):
             for name in named_params_to_update:
                 print("\t{}".format(name))
 
-        momentum=train_cfgs.get('momentum', 0.95)
+        momentum = train_cfgs.get("momentum", 0.95)
```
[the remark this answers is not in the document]
jiata: its used a few more times below

utils_cv/action_recognition/model.py
```diff
@@ -5,198 +5,140 @@
...
+        data_loaders["valid"] = self.dataset.test_dl
 
         # Move model to gpu before constructing optimizers and amp.initialize
+        device = torch_device()
         self.model.to(device)
+        count_devices = num_devices()
+        torch.backends.cudnn.benchmark = True
 
         named_params_to_update = {}
         total_params = 0
```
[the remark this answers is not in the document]
jiata: `len(self.model.named_parameters()` is a generator, so you can't use `len` on it

utils_cv/action_recognition/model.py
```diff
@@ -5,198 +5,140 @@
...
-                pin_memory=True,
-            )
+        data_loaders["train"] = self.dataset.train_dl
+        data_loaders["valid"] = self.dataset.test_dl
 
         # Move model to gpu before constructing optimizers and amp.initialize
+        device = torch_device()
         self.model.to(device)
+        count_devices = num_devices()
+        torch.backends.cudnn.benchmark = True
```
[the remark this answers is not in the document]
jiata: i believe `named_parameters` is more like `state_dict`. it gets the key/value pair.

utils_cv/action_recognition/model.py
```diff
@@ -212,17 +154,17 @@ def fit(self, train_cfgs):
...
             lr=train_cfgs.lr,
             momentum=momentum,
-            weight_decay=train_cfgs.get('weight_decay', 0.0001),
+            weight_decay=train_cfgs.get("weight_decay", 0.0001),
         )
 
         # Use mixed-precision if available
         # Currently, only O1 works with DataParallel: See issues https://github.com/NVIDIA/apex/issues/227
-        if train_cfgs.get('mixed_prec', False) and AMP_AVAILABLE:
+        if train_cfgs.get("mixed_prec", False) and AMP_AVAILABLE:
```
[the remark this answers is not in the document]
jiata: either way works. i think what you're suggesting is probably clearer tho

utils_cv/action_recognition/dataset.py
PatrickBue: Oh, is there any way around this? Training might be a lot slower with only 0 workers. And are you sure, we use >0 workers also for detection?

utils_cv/action_recognition/dataset.py
PatrickBue: Not sure if my comment was already addressed in code

utils_cv/action_recognition/dataset.py
PatrickBue: I don't think we should expose these two parameters in the function: mean and std since these are "fixed" for training, visualization, etc.

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+
+Trans = Callable[[object, dict], Tuple[object, dict]]
+
+DEFAULT_MEAN = (0.43216, 0.394666, 0.37645)
+DEFAULT_STD = (0.22803, 0.22145, 0.216989)
+
+
+class VideoRecord(object):
+    def __init__(self, row):
+        self._data = row
```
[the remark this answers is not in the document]
PatrickBue: This is marked as "hooray" but don't see a change.

utils_cv/common/misc.py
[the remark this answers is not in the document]
PatrickBue: Sorry, don't understand. Could you just switch to the standard namedtuple, or does your implementation provide extra functionality?

utils_cv/action_recognition/dataset.py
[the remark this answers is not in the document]
PatrickBue: Sounds good. Let's leave it out of this PR for now and later re-visit.

utils_cv/action_recognition/dataset.py
```diff
@@ -0,0 +1,473 @@
...
+
+        if test_split_file:
+            assert Path(test_split_file).exists()
+            assert (
+                train_split_file is not None
+                and Path(train_split_file).exists()
+            )
+
+        self.root = root
+        self.num_segments = num_segments
```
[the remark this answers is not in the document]
PatrickBue: Marked as "hooray" but don't see the edit.

utils_cv/action_recognition/model.py
```diff
@@ -5,198 +5,140 @@
...
+            tends attain higher results.
+        """
+        self.dataset = dataset
         self.model = self.init_model(
-            self.configs.sample_length,
-            self.configs.base_model,
-            self.configs.num_classes
+            self.dataset.sample_length, base_model, num_classes,
+        )
+        self.model_name = "r2plus1d_34_{}_{}".format(
```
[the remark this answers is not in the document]
PatrickBue: Marked as "hooray" but don't see change

utils_cv/action_recognition/model.py
PatrickBue: Marked as "hooray" but don't see change

utils_cv/action_recognition/model.py
PatrickBue: Marked as "hooray" but don't see change

utils_cv/action_recognition/model.py
PatrickBue: Marked as "hooray" but don't see change

utils_cv/action_recognition/dataset.py
jiata: it lkooks like i didn't upload my changes to this file... let me do that again

utils_cv/common/misc.py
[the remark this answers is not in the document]
jiata: if i switch to the standard namedtuple format, i think any lines that use this pattern: `config_object.get("field", <default_value>)`  will break

utils_cv/action_recognition/dataset.py
PatrickBue: My guess is that it will work with >0, ie setting 
   num_workers = db_num_workers()

utils_cv/action_recognition/model.py
```diff
@@ -260,11 +259,14 @@ def train_an_epoch(
                 'valid/time': ...
             }
         """
-        assert "train" in data_loaders
         if mixed_prec and not AMP_AVAILABLE:
             warnings.warn(
-                "NVIDIA apex module is not installed. Cannot use mixed-precision."
+                """
```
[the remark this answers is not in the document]
jiata: if not AMP_AVAILABLE, it'll show the warning, and  set mixed-prec to false. that ok?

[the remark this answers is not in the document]
PatrickBue: If you prefer it that way then sure.

jiata:
> Looks great, feel free to merge.
> ps. I found one probable minor bug (test for '1') and please set the number of workers in dataset db_num_workers() in a future PR (I believe this will work)

Regarding setting num_workers == db_num_workers() - this is the comment i found regarding why it is set to 0: `# Torch 1.2 has a bug when num-workers > 0 (0 means run a main-processor worker)`

As you said, in a later PR i will test out whether there is a better fix for this bug (or if the bug is even still an issue)

PR #491, opened 2020-02-10
update video annotation code

angusrtaylor:


PatrickBue:
@angusrtaylor Thanks for this PR. Just one question: is there anything left under contrib/ that is now also in scenarios/action_recognition and can hence get deleted from contrib? If so, could you perhaps do that.

angusrtaylor:
> @angusrtaylor Thanks for this PR. Just one question: is there anything left under contrib/ that is now also in scenarios/action_recognition and can hence get deleted from contrib? If so, could you perhaps do that.

I've removed the video annotation code from the contrib. Thanks

PR #497, opened 2020-02-14
Update training parameters to work with older pytorch version

PatrickBue:
### Description
Pytroch 1.2 does not support OneCycleLR scheduler, hence using the StepLR instead. But had to find good parameters, ie getting close to 80% on HMDB-51. The parameters I found for StepLR seem to work as well (almost) as using OneCycleLR.

Note: this PR does NOT clean up code or text in the notebook.

### Checklist:
- [n/a] I have added tests covering my contributions.


utils_cv/action_recognition/model.py
```diff
@@ -229,28 +233,22 @@ def fit(self, train_cfgs):
...
             )
-        elif lr_decay_steps is not None:
-            lr_decay_total_steps = train_cfgs.epochs
+        else:
             # Simple step-decay
             scheduler = torch.optim.lr_scheduler.StepLR(
                 optimizer,
-                step_size=lr_decay_steps,
-                gamma=train_cfgs.get('lr_decay_factor', 0.1),
+                step_size=train_cfgs.get('lr_step_size', None),
```
youngpark: If `lr_step_size` doesn't exist as a config parameter, it will cause failures down the road when retrieving the learning rate.

Is there a reason why we aren't falling back to a default learning rate if the config doesn't exist?

PatrickBue: Thanks for finding. None is not a valid input to step_size, so that'd be throwing an error if train_cfgs.lr_step_size isn't defined.
This parameter will reduce the learning rate by a factor of e.g. 10x, after a certain number of steps. Will edit the code so that this learning rate reduction is deactivated if train_cfgs.lr_step_size isn't provided

PR #516, opened 2020-03-08
Improve detection visualization and other additions

PatrickBue:
### Description
Improve plotting of bounding boxes e.g. with different colors, appropriate font size, or showing also the confidence.

### Checklist:


utils_cv/detection/dataset.py
```diff
@@ -282,10 +285,21 @@ def __init__(
         # create training and validation data loaders
         self.init_data_loaders()
 
+    def _verify_data(self) -> None:
```
jiata: [minor] add better description of what 'valid' means. or better yet, rename the function to be more descriptive.

utils_cv/detection/plot.py
```diff
@@ -65,26 +66,46 @@ def plot_boxes(
         draw = ImageDraw.Draw(im)
         font = get_font(size=plot_settings.text_size)
 
+        # get color map and convert to list of integer tuples
+        if not (plot_settings.rect_color and plot_settings.text_color):
```
jiata: this case should never happen since PlotSettings is initialized with a `rect_color` and a `text_color`

utils_cv/detection/plot.py
```diff
@@ -65,26 +66,46 @@ def plot_boxes(
...
                 continue
 
-            box = [(bbox.left, bbox.top), (bbox.right, bbox.bottom)]
+            # show detection score in rectangle label
+            bbox_text = bbox.label_name
+            if type(bbox) is DetectionBbox:
+                bbox_text += " ({:0.2f})".format(bbox.score)
+
+            # pick rectangle and text color
+            if plot_settings.rect_color and plot_settings.text_color:
```
jiata: i think this functionality can be part of the PlotSettings class. `plot_setting.get_next_color()` or something like that

jiata: this would also be conceptually a bit nicer because the logic of picking colors/ etc will be kept under `PlotSettings`, and the `plot_boxes` function won't have to deal with this.

utils_cv/detection/dataset.py
PatrickBue: @jiata Thanks! Will address your comments and then re-request review. Also added some new visualization functionality today, which I need to write tests for before the next review.

PatrickBue: Added more text to hopefully make this more clear.

utils_cv/detection/plot.py
PatrickBue: Added get_color() function to the PlotSettings class. Does indeed look nicer

utils_cv/detection/plot.py
PatrickBue: i implemented it such that if plot_settings.rect_color is set to None, it instead cycles through all colors.

tests/unit/detection/test_detection_dataset.py
```diff
@@ -221,6 +223,40 @@ def test_detection_dataset_init_train_pct(
     assert len(data.train_ds) == 24
 
 
+def test_detection_dataset_verify(basic_detection_dataset):
```
jiata: ideally this unit test should check that the warning is called if image resolution is too high, but i think this is ok for now

PR #523, opened 2020-03-18
Adding coco to voc conversion script.

PatrickBue:
### Description
Adding functionality to convert from COCO to VOC. 

### Checklist:


environment.yml
```diff
@@ -36,6 +36,7 @@ dependencies:
 - pre-commit>=1.14.4
 - pyyaml>=5.1.2
 - requests>=2.22.0
+- cytoolz
```
PatrickBue: @jiata: I added the 04 notebook to get some early feedback. It's unfinished, but I am planning to clean it up a little and correct typos and would then also submit to staging (even without the speed/accuracy tests done, but showing that we can indeed reproduce the mAP number from torchvision)

PatrickBue: ps. leaving a comment here since I cannot leave a comment in the 04 notebook itself (too big for github)

data/misc/coco/coco_sample.json
```diff
@@ -0,0 +1,120 @@
...
+    ],
+    "images": [
+        {
+            "id": 0,
+            "license": 1,
+            "file_name": "image0.jpg",
+            "height": 480,
+            "width": 640,
+            "date_captured": "2020-01-01T00:00:00+00:00",
+			"coco_url": "http://image0.jpg"
```
jiata: this formatting is weird

tests/conftest.py
```diff
@@ -683,6 +683,16 @@ def od_detections(od_detection_dataset):
     return learner.predict_dl(od_detection_dataset.test_dl, threshold=0)
 
 
+@pytest.fixture(scope="session")
+def coco_sample_path(tmp_session) -> str:
+    """ Returns the path to a coco-formatted annotation. """
+    return os.path.abspath(
+        os.path.join(
+            os.path.dirname(__file__), os.path.pardir, "data", "misc", "coco", "coco_sample.json"
```
jiata: really really minor, but i think we've been using this convention elsewhere:
```
os.path.dirname(__file__), os.path.pardir, "data/misc/coco/coco_sample.json"
```

utils_cv/detection/references/anno_coco2voc.py
```diff
@@ -0,0 +1,214 @@
+# Code copied and slightly modified from:
+# https://github.com/CasiaFan/Dataset_to_VOC_converter/blob/master/anno_coco2voc.py
+#
+# Most modifications are hlighlighted by the keyword "EDITED".
```
jiata: you probably did this, but best to make sure *all* modifications are highlighted with EDITED

jiata:
<p>i didn't look with too much detail bc it looks like this notebook will still be edited(?) but this is great. really cool to see this analysis with coco dataset.</p>

data/misc/coco/coco_sample.json
```diff
@@ -0,0 +1,120 @@
+{
```
jiata: shouldn't we not be saving anything under the `data` folder?  i think `data` is ignored in our gitignore (and if its not, it probably should be)

PatrickBue: Yeah, was wondering about that also, what to download, and what to put into the repo. So far, do we ever download anything other than images? We of course already have the data/misc folder, but guess all that's in there so far is NOT for testing but for user to run the notebook.

data/misc/coco/coco_sample.json
PatrickBue: Yes. Looks like this extra tab is coming from github, it looks good in a text editor.

tests/conftest.py
PatrickBue: In that case, maybe we leave my code as is, since using os.path.join seems like the right thing to do rather than "/"s ?

utils_cv/detection/references/anno_coco2voc.py
PatrickBue: Let me double-check, but yes, that was my intention

PatrickBue:
@jiata I finished the 04 notebook and addressed you questions - with questions of my own mostly :)

data/misc/coco/coco_sample.json
PatrickBue: An alternative (which I sort of like best of all options) could be that I create a new file ./tests/resources.py and copy the content of the .json file there surrounded by """ quotes, import that in conftest.py, and then just write it to a temporary file during testing

jiata: i prefer the second thing you suggested. actually what we currently have under `data/misc` (which is only used for action rec) should be downloaded from blob and not saved directly in the repo.

jiata: in a later PR for action rec, i'll likely delete the files inside `data/misc`

data/misc/coco/coco_sample.json
youngpark: [nit] although its hard to see, I think there actually is a tab instead of spaces here.

data/misc/coco/coco_sample.json
youngpark: I think uploading the `data/*` to a blob makes much sense as I remember seeing a thread on the OS AI recommenders channel that compared the sizes of our repositories and this will help reduce the size of our repo.
`272M    computervision-recipes`
`162M    forecasting`
`50M     nlp-recipes`
`33M     recommenders`

data/misc/coco/coco_sample.json
PatrickBue: Thanks. I deleted the file and also ran black on the xml string.

data/misc/coco/coco_sample.json
PatrickBue: Removed the file and instead added string as suggested above.

jiata: @youngpark yikes, didn't realize ours was so big

PatrickBue:
@jiata Regarding "update the dataset object itself to accept coco-annotations": totally agree. When we have an AI lab on object detection then could re-visit, based on the script I added in this PR. This could be a bit more work if we also want to support mask and keypoint annotations. For now I wanted to mainly unblock Raj.

PatrickBue:
@jiata ps, tried to run black on the cell, but it splits it up onto way too many lines.

PR #530, opened 2020-04-06
Image similarity widget

PatrickBue:
### Description
Added interactive widget for image similarity

### Checklist:


PatrickBue:
@jiata can you please review

utils_cv/similarity/widget.py
```diff
@@ -14,160 +15,131 @@ def _list_sort(list1D, reverse=False, comparison_fct=lambda x: x):
...
-                    children=[
-                        w_header,
-                        widgets.HBox(children=[self.w_img, w_info]),
-                    ]
-                )
-            ]
-        )
-        self.ui.set_title(0, "Results viewer")
+        # Create tab
+        self.ui = widgets.Tab(children=[w_img_grid])
```
jiata, line 141: does this need to be a tab?

PatrickBue: No, but it looks nicer with tab :)

PatrickBue:
@jiata thanks for the feedback. Agree that the yellow button is not ideal, much better would be if we could click on the image itself (but the image widget class doesn't have a click callback). I'll leave it as is for now, and maybe leave it for later should we ever want to improve the UI.

PR #550, opened 2020-06-16
Plot changes

sharatsc:
Adding option to override text size, rectangle and keypoint thickness

### Description

Allows option to override annotation properties such as text size, rectangle thickness and keypoint thickness

The default text size as not always suitable

### Checklist:

- [X] I have updated the documentation accordingly.
- [X] This PR is being made to `staging` and not `master`
- [X] I will squash merge this PR into `staging`


utils_cv/detection/plot.py
```diff
@@ -268,6 +268,9 @@ def plot_detections(
     idx: int = None,
     keypoint_meta: Dict = None,
     ax: plt.axes = None,
+    text_size: int = None,
```
PatrickBue, line 271: Instead of the 3 variables, should this take a PlotSettings() object as input?

sharatsc: The function overrides plot settings multiple places with hard coded settings (separate colors for GT vs. detection, separate colors for keypoint vs boxes). These are the only default settings. These are the default settings that are relevant for smaller/larger image annotations

PR #546, opened 2020-06-09
utils_cv for mot

karineip:
### Description
Added utility functions called in training notebook and TrackingLearner evaluate() function. Test and raised errors will be added later.

- In  in file_format.py: convert_vott_MOTxywh()
- In display_with_bb.py:  convert_trackingbboxes_txt(), convert_trackingbboxes_video()

### Checklist:

- [ ] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ ] I have added tests covering my contributions.
- [ ] I have updated the documentation accordingly.
- [ ] This PR is being made to `staging` and not `master`
- [ ] I will squash merge this PR into `staging`

PR #553, opened 2020-06-16
FairMOT-01

caseyhong:
### Description

- updated the 01_training_introduction notebook to include the evaluation and prediction examples
- debugged the `predict` method of `TrackingLearner`
- added a small amount of logging in the `fit` method of `TrackingLearner`
- used `unzip_url` and `download` utils instead of hard-coded datapaths
- added a separate `load` method to `TrackingLearner` to specify paths to model weights
- added basic plotting functionality
- set the logging setting to WARNING instead of DEBUG in the source code (and marked as EDITED in the readme)

### Checklist:


scenarios/tracking/01_training_introduction.ipynb
```diff
@@ -20,7 +20,7 @@
    "cell_type": "markdown",
```
PatrickBue, line 20: <ul><li>Can we use the variable name IM_SIZE = (1920, 1080) to be consistent with other scenarios</li><li>Is TRACK_BUFFER an important parameter? If not, can we just set+use the 300 as default in the respective function.</li><li>Looks like FRAME_RATE isn't used and can be deleted. Also not sure BASELINE_MODEL and FT_MODEL are needed.</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/computervision-recipes/pull/553/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

PatrickBue, line 20: <p>Is this safe and load needed, or here mainly as an example? If example, then could comment out the code and mention that this is an example to save/load a model. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/computervision-recipes/pull/553/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

utils_cv/tracking/data.py
```diff
@@ -0,0 +1,4 @@
+class Urls:
```
PatrickBue, line 1: Can you please have a look at e.g. object detection and copy how the class looks there including the all() function

utils_cv/tracking/plot.py
```diff
@@ -0,0 +1,156 @@
...
+        for _, v in preds.items()
+        for bb in v
+    ]
+
+    df = pd.DataFrame(
+        bboxes, columns=["frame", "id", "left", "top", "width", "height",],
+    )
+
+    data_dir, video_name = osp.split(video_path)
+    output_path = osp.join(data_dir, f"track_{video_name}")
```
PatrickBue, line 47: [minor] output_path should probably be a input parameter to this function.

scenarios/tracking/01_training_introduction.ipynb
caseyhong: BASELINE_MODEL and FT_MODEL are required for `load`

utils_cv/tracking/plot.py
```diff
@@ -0,0 +1,156 @@
...
+    _write_video(df, video_path, output_path, frame_rate)
+    print(f"Output saved to {output_path}.")
+
+    return output_path
+
+
+def _write_video(
+    df: pd.DataFrame, input_video: str, output_video: str, frame_rate: int
```
PatrickBue, line 55: Was this function/code copied from somewhere? If not, do we need to use pandas here or could this simply use whatever predict outputs (guess results: Dict[int, List[TrackingBbox]])

utils_cv/tracking/plot.py
```diff
@@ -0,0 +1,156 @@
...
+import numpy as np
+import pandas as pd
+
+from .bbox import TrackingBbox
+
+
+def plot_results(
+    results: Dict[int, List[TrackingBbox]],
+    video_path: str,
+    frame_rate: int = 30,
```
PatrickBue, line 14: Should the frame rate here simply be the same frame rate as the video? Maybe there is something like this:
frame_rate = int(video.get(cv2.FRAME_RATE))

utils_cv/tracking/plot.py
```diff
@@ -0,0 +1,156 @@
...
+
+def _xywh_to_xyxy(xywh: List[int]) -> List[int]:
+    """ Convert bbox of form (left,top,width,height) to (left,top,right,bottom) """
+    left, top, width, height = xywh
+    return [left, top, left + width, top + height]
+
+
+def _draw_boxes(
+    im: np.ndarray,
+    cur_df: pd.DataFrame,
```
PatrickBue, line 98: Same here, can this not use pandas so that the function can more easily be re-used?

scenarios/tracking/01_training_introduction.ipynb
caseyhong: <p>Mainly as an example - but also because if you just run in straight through, we only train for 1 epoch so the subsequent <code>predict</code>  runs would output basically no tracks. This is why I show how to load weights from a model Karine trained  for 30 epochs before calling <code>predict.</code> I can wrap the <code>save</code> method in <code>if SAVE_MODEL = True</code>, but what do you think would be the best way to handle the <code>load</code> examples? </p>

utils_cv/tracking/model.py
```diff
@@ -159,34 +148,63 @@ def fit(
         if not self.dataset:
             raise Exception("No dataset provided")
 
-        self.opt.lr = lr
-        self.opt.lr_step = lr_step
-        self.opt.num_epochs = num_epochs
+        opt_fit = deepcopy(self.opt)
```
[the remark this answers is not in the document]
caseyhong: It leads to bugs later when you use the `TrackingLearner` to do `predict` 😢 (e.g. for predict, you need "fresh" `opt.<dataset_params>`

utils_cv/tracking/model.py
```diff
@@ -112,33 +119,15 @@ def init_model(self) -> nn.Module:
...
-                resume,
-                self.opt.lr,
-                self.opt.lr_step,
-            )
-        else:
-            # otherwise just load the model state dict
-            self.model = load_model(self.model, path)
+    def load(self, path, resume=False) -> None:
+        self.opt.load_model = path
+        self.opt.resume = resume
```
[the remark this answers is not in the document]
caseyhong: Yeah it's because we actually call FairMOT's `load_model` at different steps, just need to update `opt`

utils_cv/tracking/model.py
```diff
@@ -2,29 +2,34 @@
 # Licensed under the MIT License.
 
 import argparse
+from copy import deepcopy
 import os
 import os.path as osp
 from typing import Dict, List
 import requests
+import glob
```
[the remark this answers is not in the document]
caseyhong: It's used in `_get_dataloader`

utils_cv/tracking/plot.py
caseyhong: It's from [https://github.com/microsoft/computervision-recipes/pull/546](https://github.com/microsoft/computervision-recipes/pull/546)

caseyhong: I can refactor tho

utils_cv/tracking/model.py
```diff
@@ -159,34 +148,63 @@ def fit(
...
-        self.optimizer = torch.optim.Adam(self.model.parameters(), self.opt.lr)
-        self.start_epoch = 0
-
-        Trainer = train_factory[self.opt.task]
-        trainer = Trainer(self.opt.opt, self.model, self.optimizer)
-        trainer.set_device(
-            self.opt.gpus, self.opt.chunk_sizes, self.opt.device
+        self.optimizer = torch.optim.Adam(self.model.parameters(), opt_fit.lr)
+        start_epoch = 0
+        self.model, self.optimizer, start_epoch = load_model(
```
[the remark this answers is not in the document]
caseyhong: - I don't think there is native support for this but it shouldn't be too hard to add our own `resume` param to `fit` and only `load_model` if `resume` is False. I would prefer to add that after some tests are in place tho if that's alright
- At initialization, self.model just has the architecture (which is why we need to know if it is dla_34 or hrnet_32, etc)

utils_cv/tracking/model.py
```diff
@@ -218,35 +236,38 @@ def predict(
...
+        opt_pred.conf_thres = conf_thres
+        opt_pred.det_thres = det_thres
+        opt_pred.nms_thres = nms_thres
+        opt_pred.track_buffer = track_buffer
+        opt_pred.min_box_area = min_box_area
 
         input_height = input_h if input_h else -1
         input_width = input_w if input_w else -1
-        self.opt.update_dataset_res(input_height, input_width)
+        opt_pred.update_dataset_res(input_height, input_width)
```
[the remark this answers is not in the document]
caseyhong: It's because updating `input_h` and `input_w` is special because it affects some other params in `opts`, e.g. `output_h` and `output_w`

scenarios/tracking/01_training_introduction.ipynb
PatrickBue: Sounds all good. We might remove TRACK_BUFFER later then, if it is overly specific for our toy dataset.

PatrickBue: Sounds like we can leave this as is for now since it helps us speed up model development (due to us not wanting to wait 30 epochs every time we run the notebook). Longer term I would suggest to just comment out the two lines and write a comment that this is how one would load/save the model.

utils_cv/tracking/model.py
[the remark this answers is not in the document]
PatrickBue: Get it. Could you perhaps add a quick comment such as 
opt_fit = deepcopy(self.opt) #copy opt to avoid bug

utils_cv/tracking/model.py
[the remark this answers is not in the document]
PatrickBue: Don't fully understand. But seems odd that there's a "load()" function which doesn't really load the model but only updates some opts.

utils_cv/tracking/model.py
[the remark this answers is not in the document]
PatrickBue: - Sounds good, let's leave as is and only add the "resume" functionality later. That way we also avoid adding bugs which would later be really hard to find.
- I am a bit confused. There is a self.model which the user can specify in the constructor, and it's overwritten here somehow.

utils_cv/tracking/model.py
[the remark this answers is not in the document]
PatrickBue: Makes sense.

utils_cv/tracking/plot.py
PatrickBue: Yeah, please remove all dependencies on Pandas so that it can be used easily and directly with output from the model

PatrickBue:
@caseyhong I just left some quick feedback to your comments. Once you want me to do another iteration again, just press the "re-request review" button

scenarios/tracking/01_training_introduction.ipynb
```diff
@@ -236,22 +228,23 @@
...
+      "off_loss: 0.18636367223715702\n",
+      "id_loss: 0.8860541224528692\n",
+      "time: 44.016666666666666\n",
+      "Model saved to ./models/fairmot_ft.pth\n"
      ]
     }
    ],
    "source": [
-    "tracker.fit(num_epochs=EPOCHS, lr=LEARNING_RATE)"
+    "tracker.fit(num_epochs=EPOCHS, lr=LEARNING_RATE, resume=True)"
```
PatrickBue, line 242: This resume=True seems odd since we are starting to train on a new dataset.

utils_cv/tracking/model.py
```diff
@@ -99,7 +99,7 @@ def __init__(
 
         Args:
             dataset: the dataset
-            model: the model
+            model: path to save model
```
PatrickBue, line 102: [minor] should be rename to model_path if it's a path

utils_cv/tracking/model.py
```diff
@@ -192,6 +193,8 @@ def fit(
                 for param_group in optimizer.param_groups:
                     param_group["lr"] = lr
 
+        self.save(self.model_path)
```
PatrickBue, line 196: Can you describe here a bit more why we save (because the original functionality assumes the model being written to disk etc)

utils_cv/tracking/plot.py
```diff
@@ -24,39 +22,24 @@ def plot_results(
...
-            bb.right - bb.left,
-            bb.bottom - bb.top,
-        ]
-        for _, v in preds.items()
-        for bb in v
-    ]
-
-    df = pd.DataFrame(
-        bboxes, columns=["frame", "id", "left", "top", "width", "height",],
+    _write_video(
```
PatrickBue, line 26: Is this "_write_video" left-over code that should get deleted?

utils_cv/tracking/plot.py
```diff
@@ -24,39 +22,24 @@ def plot_results(
...
-    print(f"Output saved to {output_path}.")
+    print(f"Output saved to {output_video}.")
 
-    return output_path
+    return output_video
 
 
 def _write_video(
-    df: pd.DataFrame, input_video: str, output_video: str
+    results: Dict[int, List[TrackingBbox]], input_video: str, output_video: str
```
PatrickBue, line 38: Should this function be called write_video() (without the "_") and be moved to model.py ?

utils_cv/tracking/plot.py
```diff
@@ -68,53 +51,51 @@ def _write_video(
...
     """
 
-    cur_ids = cur_df.id.tolist()
+    cur_ids = [bb.track_id for bb in cur_tracks]
+    tracks = dict(zip(cur_ids, cur_tracks))
 
-    for i, id in enumerate(cur_ids):
-        xywh = [round(x) for x in cur_df.loc[cur_df.id == id].values[0][2:]]
-        left, top, right, bottom = _xywh_to_xyxy(xywh)
+    for tid, bb in tracks.items():
```
PatrickBue, line 90: [minor] would label be a better name than tid?

scenarios/tracking/01_training_introduction.ipynb
caseyhong: Because we default to finetuning the all_dla34 model, which is frozen at epoch 10

PR #566, opened 2020-06-24
Kaipkiun/01 evaluate plotloss

karineip:
### Description
Changes include:
- 01 notebook updated to run `plot_training_loss` and show `evaluate` fn output
- `evaluate` fn added to model.py
- `plot_training_loss` fn added to model.py
- `save_results_txt` added to plot.py
- Fairmot scripts evaluation.py and io.py updated to fix import

Request feedback on:
-  as commented in PR 558, if `plot_training_loss`  needs to be moved to plot.py, and if ok to change fit to return `loss_dict`
-  how to format output of evaluate fn, currently spitting typical pymotmetrics metrics.

### Checklist:

### DIFF
--- a/utils_cv/tracking/model.py
+++ b/utils_cv/tracking/model.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 import argparse
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from copy import deepcopy
 import glob
 import requests
@@ -18,6 +18,7 @@ from torch.utils.data import DataLoader
 import cv2
 import pandas as pd
 import matplotlib.pyplot as plt
+import motmetrics as mm
 
 from .references.fairmot.datasets.dataset.jde import LoadImages, LoadVideo
 from .references.fairmot.models.model import (
@@ -26,12 +27,13 @@ from .references.fairmot.models.model import (
     save_model,
 )
 from .references.fairmot.tracker.multitracker import JDETracker
+from .references.fairmot.tracking_utils.evaluation import Evaluator
 from .references.fairmot.trains.train_factory import train_factory
 
 from .bbox import TrackingBbox
 from .dataset import TrackingDataset
 from .opts import opts
-from .plot import draw_boxes, assign_colors
+from .plot import draw_boxes, assign_colors, save_results_txt
 from ..common.gpu import torch_device
 
 BASELINE_URL = (
@@ -211,7 +213,10 @@ class TrackingLearner(object):
         Trainer = train_factory[opt_fit.task]
         trainer = Trainer(opt_fit.opt, self.model, self.optimizer)
         trainer.set_device(opt_fit.gpus, opt_fit.chunk_sizes, opt_fit.device)
-
+        
+        # initialize loss vars
+        self.losses_dict = defaultdict(list)
+        
         # training loop
         for epoch in range(
             start_epoch + 1, start_epoch + opt_fit.num_epochs + 1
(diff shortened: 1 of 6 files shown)


utils_cv/tracking/model.py
```diff
@@ -243,9 +279,48 @@ def save(self, path) -> None:
...
+            strsummary: pandas.DataFrame output by method in 'motmetrics', containing metrics scores
+        Raises:
+            Exception: if both `results` and `self.stored_predictions` are None.
+        """
+        if results is None:
+            if not self.stored_predictions: #TODO: add stored_predictions in predict() 
+                raise Exception("No predict() function run on dataset for for evaluation")
+            results = self.stored_predictions     
+        
+        result_filename = osp.join(gt_root_path,'results', 'results.txt')
```
PatrickBue, line 300: Instead of writing this file to the user directory, use tempfile.TemporaryDirectory() instead. You can then create a subdirectory called "results" there which the user will never see.

utils_cv/tracking/model.py
```diff
@@ -243,9 +279,48 @@ def save(self, path) -> None:
         save_model(path, self.epoch, self.model, self.optimizer)
         print(f"Model saved to {path}")
 
-    def evaluate(self, results, gt) -> pd.DataFrame:
-        pass
+    def evaluate(self,
+                 results: Dict[int, List[TrackingBbox]],
+                 gt_root_path: str) -> pd.DataFrame:
```
PatrickBue, line 284: Don't think this ground truth root path is needed if we use tempfile.TemporaryDirectory(). See below.

utils_cv/tracking/model.py
```diff
@@ -243,9 +279,48 @@ def save(self, path) -> None:
...
+        """ eval code that calls on 'motmetrics' package in referenced FairMOT script, to produce MOT metrics on inference, given ground-truth.
+        Args:
+            results: prediction results from predict() function, i.e. Dict[int, List[TrackingBbox]] 
+            gt_root_path: path of dataset containing GT annotations in MOTchallenge format (xywh)
+        Returns:
+            strsummary: pandas.DataFrame output by method in 'motmetrics', containing metrics scores
+        Raises:
+            Exception: if both `results` and `self.stored_predictions` are None.
+        """
+        if results is None:
```
PatrickBue, line 295: What is the purpose of "self.stored_predictions" and can it get deleted? Then this section "if results" would not be necessary anymore.

utils_cv/tracking/model.py
```diff
@@ -243,9 +279,48 @@ def save(self, path) -> None:
...
+        mh = mm.metrics.create()
+       
+        summary = Evaluator.get_summary(accs, ("single_vid",), metrics)
+        strsummary = mm.io.render_summary(
+            summary,
+            formatters=mh.formatters,
+            namemap=mm.io.motchallenge_metric_names
+        )
+        print(strsummary)
+        Evaluator.save_summary(summary, osp.join(gt_root_path,'results', 'summary_metrics.xlsx'))
```
PatrickBue, line 320: No need to save the results to disk.

utils_cv/tracking/model.py
```diff
@@ -243,9 +279,48 @@ def save(self, path) -> None:
...
+            results = self.stored_predictions     
+        
+        result_filename = osp.join(gt_root_path,'results', 'results.txt')
+        if not osp.exists(osp.join(gt_root_path,'results')):
+            os.makedirs(osp.join(gt_root_path,'results'))
+        save_results_txt(results, result_filename)
+
+        #Implementation inspired from code found here: https://github.com/ifzhang/FairMOT/blob/master/src/track.py
+        evaluator = Evaluator(gt_root_path, "single_vid", "mot")
+        accs=[evaluator.eval_file(result_filename)]
```
catherine667, line 307: Could you separate this evaluate from here? take accs as an input for another function maybe called eval_summary(). The following code should take a list of result_filenames.

utils_cv/tracking/model.py
```diff
@@ -243,9 +279,48 @@ def save(self, path) -> None:
...
+                raise Exception("No predict() function run on dataset for for evaluation")
+            results = self.stored_predictions     
+        
+        result_filename = osp.join(gt_root_path,'results', 'results.txt')
+        if not osp.exists(osp.join(gt_root_path,'results')):
+            os.makedirs(osp.join(gt_root_path,'results'))
+        save_results_txt(results, result_filename)
+
+        #Implementation inspired from code found here: https://github.com/ifzhang/FairMOT/blob/master/src/track.py
+        evaluator = Evaluator(gt_root_path, "single_vid", "mot")
```
catherine667, line 306: Make "single_vid" a parameter to be passed in.

utils_cv/tracking/plot.py
```diff
@@ -1,13 +1,44 @@
...
 
+from collections import OrderedDict
 import os.path as osp
 from typing import Dict, List, Tuple
 import cv2
 import numpy as np
 
 from .bbox import TrackingBbox
 
+def save_results_txt(
```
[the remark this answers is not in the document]
karineip: Ok sounds good. Would it be better to name it boxes_to_mot()? if mot means mot challenge format.

utils_cv/tracking/model.py
karineip: @catherine667, can you clarify more what you mean and why?

utils_cv/tracking/plot.py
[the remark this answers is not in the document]
PatrickBue: Good point, it should be "boxes_to_mot()"

utils_cv/tracking/model.py
PatrickBue: I don't follow either.

utils_cv/tracking/model.py
karineip: Good point, i was doing it like in detection/model.py. But the context is different because it could use the predictions from test dataset stored internally. Will remove

utils_cv/tracking/model.py
karineip: It's still used by ` evaluator = Evaluator(gt_root_path, "single_vid", "mot")` to locate the gt annotation file. I'll use tempfile.TemporaryDirectory() to write the transient `results.txt` file, before `results.txt` file was output into this groundtruth root path too.

PR #571, opened 2020-06-26
FairMOT patch

caseyhong:
### Description

🐛 fixes
- set track_id to 0-index for predict()
- fix bounding box visualization errors
- remove non-functional code (downloading google drive weights)
- fix small bugs in the notebook code

### Checklist:


utils_cv/tracking/references/fairmot/tracker/multitracker.py
```diff
@@ -190,6 +189,7 @@ def __init__(self, opt, frame_rate=30):
         self.std = np.array(opt.std, dtype=np.float32).reshape(1, 1, 3)
 
         self.kalman_filter = KalmanFilter()
+        BaseTrack._count = 0 # EDITED
```
PatrickBue, line 192: What does this do / why is this needed?

utils_cv/tracking/references/fairmot/datasets/dataset/jde.py
```diff
@@ -94,8 +94,8 @@ def __init__(self, path, img_size=(1088, 608)):
         self.height = img_size[1]
         self.count = 0
 
-        self.w, self.h = self.width, self.height # EDITED
-        print('Lenth of the video: {:d} frames'.format(self.vn))
+#         self.w, self.h = 1920, 1080
```
PatrickBue, line 97: [minor] can you add an #EDITED here please

utils_cv/tracking/model.py
```diff
@@ -39,48 +38,6 @@
 from .plot import draw_boxes, assign_colors
 from ..common.gpu import torch_device
 
-BASELINE_URL = (
-    "https://drive.google.com/open?id=1udpOPum8fJdoEQm6n0jsIgMMViOMFinu"
-)
-
-
-def _download_baseline(url, destination) -> None:
```
PatrickBue, line 47: Is this function not needed anymore?

utils_cv/tracking/model.py
```diff
@@ -163,18 +120,23 @@ def __init__(
...
     def init_model(self) -> nn.Module:
         """
         Download and initialize the baseline FairMOT model.
         """
-        model_dir = osp.join(self.opt.root_dir, "models")
-        baseline_path = osp.join(model_dir, "all_dla34.pth")
-        #         os.makedirs(model_dir, exist_ok=True)
-        #         _download_baseline(BASELINE_URL, baseline_path)
-        self.opt.load_model = baseline_path
+        if osp.isfile(self.model_path):
```
PatrickBue, line 130: Could rename function to "_init_model()" to indicate it's only used internally, ie the user will never call it.

utils_cv/tracking/model.py
```diff
@@ -163,18 +120,23 @@ def __init__(
...
+            self.opt.load_model = self.model_path
+        else:
+            baseline_path = osp.join(
```
PatrickBue, line 133: This is a bit funny. If "model_path" is a file then use it, otherwise it gets totally ignored. Maybe instead check if "model_path" is None?

utils_cv/tracking/model.py
```diff
@@ -282,46 +242,46 @@ def save(self, path) -> None:
...
         summary = Evaluator.get_summary(accs, ("single_vid",), metrics)
         strsummary = mm.io.render_summary(
             summary,
             formatters=mh.formatters,
-            namemap=mm.io.motchallenge_metric_names
+            namemap=mm.io.motchallenge_metric_names,
         )
-        print(strsummary)        
-        
+        print(strsummary)
```
PatrickBue, line 281: [minor] Ideally this print shouldn't be here but in the notebook.

scenarios/tracking/01_training_introduction.ipynb
```diff
@@ -25,7 +25,7 @@
   },
```
PatrickBue, line 25: <p>Do we need the FRAME_RATE here, and can we also remove TRACK_BUFFER?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/computervision-recipes/pull/571/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

utils_cv/tracking/model.py
karineip: Ok, Liqun fixed that in her PR.

utils_cv/tracking/references/fairmot/tracker/multitracker.py
caseyhong: So before this was added, if you had the notebook open and you ran predict() multiple times (let's say twice, for this example), the following would happen: 
first run:
predict() -> 3 tracks with track_ids: [0, 1, 2]
second run:
predict() -> 4 tracks with track_ids: [3, 4, 5, 6]

This is not the desired result because we could be running predict() on two different datasets, just with the same TrackingLearner object.

This was happening because of how the `nextid` method was implemented in BaseTrack in the source code; I guess because the code was designed to be run on the command line.

utils_cv/tracking/model.py
caseyhong: It never really worked lol because I couldn't find the url to directly download

utils_cv/tracking/model.py
caseyhong: The thinking behind this was that a user may want to initialize a TrackingLearner only to use for inference. In this case they could supply `model_path` (an existing file) as a param. Otherwise, we default to `all_dla34.pth` but still ask for `model_path` to be specified so that we can save the finetuned weights to that path when `fit` is called. (Recall that `predict` currently depends on the src `JDETracker`, which must read the model weights from disk). 

I agree it is a bit weird, am open to suggestions.

utils_cv/tracking/references/fairmot/tracker/multitracker.py
PatrickBue: Great, thanks for the explanation.

utils_cv/tracking/model.py
PatrickBue: OK for now, but def weird to over-write a passed in model (if I understand correctly). How about as discussed we add the functionality to load a model in JDETracker by allowing a model object to be passed in (I realize this means we slightly edit the original FairMOT code). That way we don't need to save things anymore in fit, and also we can use/wrap the existing model loading functionality in FairMOT

PR #577, opened 2020-07-02
Lorenzov readme

LorenzoMicrosoft:
### Checklist:

- [x ] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [NA ] I have added tests covering my contributions.
- [x ] I have updated the documentation accordingly.
- [x ] I will squash merge this PR into `staging`

### DIFF
--- a/scenarios/tracking/FAQ.md
+++ b/scenarios/tracking/FAQ.md
@@ -9,7 +9,6 @@
 This document includes answers and information relating to common questions and topics regarding multi-object tracking. For more general Machine Learning questions, such as "How many training examples do I need?" or "How to monitor GPU usage during training?", see also the image classification [FAQ](https://github.com/microsoft/ComputerVision/blob/master/classification/FAQ.md).
 
 * General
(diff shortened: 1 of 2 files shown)


LorenzoMicrosoft:
> This looks great, and I like how you de-cluttered the readme.md file. Could we move perhaps a bit of content back from the FAQ just so that the user understand from a quick glance that we did a good amount of literature review and we decided to use an approach (FairMOT) which works really well in comparison.

We could definitely do that if you want. I think moving the "Why FairMOT" section should be enough. What do you think?

PatrickBue:
> > This looks great, and I like how you de-cluttered the readme.md file. Could we move perhaps a bit of content back from the FAQ just so that the user understand from a quick glance that we did a good amount of literature review and we decided to use an approach (FairMOT) which works really well in comparison.
> 
> We could definitely do that if you want. I think moving the "Why FairMOT" section should be enough. What do you think?

Perfect

PR #598, opened 2020-07-13
Automatic FairMOT baseline model download and MOT17 download patch

abuendia:
### Description

This PR adds:
1. Automatic FairMOT model download in 01 and 02 tracking notebooks. The baseline dla model is uploaded to the public cvbp blob container and gets pulled down into both notebooks. Removes the need for the user to manually download the model from a Google Drive link.
2. Automatic download of MOT17 into 02 notebook. The relevant train/test subsets of MOT17 (2 GB) are uploaded to the public cvbp blob container and get pulled into the 02 notebook. I extract only the subdirectories needed for 02 to avoid a 5 GB download.
3. Minor changes to 01 and 02 markdown, i.e. typo/grammar fixes.

Reuses unzip_url util. Downloads if does not exist, and avoids re-downloading if exists. Adds relevant download URLs to `data.py`.

**Important:** When running the current version of 02 in staging, I encountered a bug that indicates the MOT17 zip from the MOT Challenge website was recently changed and corrupted. I have uploaded an earlier working version from @karineip to blob store.

### Checklist:


utils_cv/tracking/data.py
```diff
@@ -6,11 +6,14 @@
...
+    datasets_base = "https://cvbp.blob.core.windows.net/public/datasets/tracking/"
+    models_base = "https://cvbp.blob.core.windows.net/public/models/tracking/"
 
-    cans_path = urljoin(base, "cans.zip")
-    fridge_objects_path = urljoin(base, "odFridgeObjects_FairMOT-Format.zip")
-    carcans_annotations_path = urljoin(base, "carcans_vott-csv-export.zip")
+    cans_path = urljoin(datasets_base, "cans.zip")
+    fridge_objects_path = urljoin(datasets_base, "odFridgeObjects_FairMOT-Format.zip")
+    carcans_annotations_path = urljoin(datasets_base, "carcans_vott-csv-export.zip")
+    mot_challenge_path = urljoin(datasets_base, "MOT17.zip")
```
PatrickBue, line 15: We aren't allowed to host data we don't own ourselves. Can you download from the original non-microsoft location?

abuendia: Thanks - unfortunately I don't think so. I made this change because I tried running the 02 notebook currently in staging and got an error, even using a fresh VM and clone of the repo. The zip that gets downloaded from MOTChallenge website is corrupted for me. I committed the error message here: https://github.com/microsoft/computervision-recipes/blob/c0078e2c5490777ae6e90fc4073d49bfecd28582/scenarios/tracking/02_mot_challenge.ipynb

Do you mind trying a run of 02 in staging just to be sure it's not me?

abuendia: Would this comment also apply to the baseline dla model?

PatrickBue: I also got the same error like you before with the zip file

abuendia: Also cc'ing @catherine667 and @karineip in case they are able to run the current staging 02 without a bug.

abuendia: @PatrickBue Ah ok, thanks! Here's a suggestion - I can try using a different year from MOTChallenge website and see if it works. What do you think?

PatrickBue: If that's quick and easy to try then yeah (please @catherine667 let us know if that's not an option).

Generally I wouldn't spend much more time on this since the MAIDAP project officially ended :)

abuendia: No problem, I think it should be really fast to make those edits. Just need to rerun and replace the markdown numbers at the final comparison. I'll give it a try tonight and push updates here. Do you think it's ok to host the baseline FairMOT model in blob for automatic download?

PatrickBue: Good point. We'd have to check with CELA so better not

abuendia: Sounds good. So I'll:
1. Revert the model download back to the Google drive link. The user will have to manually click and download but I think that's the only completely compliant way. 
2. Make 02 use a different year.

PatrickBue:
@abuendia can this PR be closed?

